PR #40, opened 2023-10-02
Add cmake build for Mac Intel without Metal

dotieuthien:
- Error when loading llama model with Metal on MacBook with Intel, so I fix the document to run nitro_temp with only-CPU mode.
- Configuration: Intel Core i7 and Intel Iris Plus


tikikun (approved):
Metal is not working on mac intel now, approved

Issue #69, opened 2023-10-12
feat: support http cors 


nam-john-ho:
**Problem**
User could not call llama/chat_completion api on web browsers due to CORS policy
![Screenshot 2023-10-12 at 12 21 15](https://github.com/janhq/nitro/assets/146448941/b20969c4-84d0-4df0-8d44-73f9f17c2048)

**Success Criteria**
Users able to call llama/chat_completion api on web browsers

PR #72, opened 2023-10-12
#89: support cors chat completion api

nam-john-ho:
https://github.com/janhq/nitro/issues/69


tikikun:
image

According to drogon it seems that the default behaviour is having cors enabled

PR #92, opened 2023-10-25
chore: Amend license from MIT to AGPLv3

dan-jan:
Folks, I've submitted a PR to shift Nitro from MIT license to AGPL, as per our retreat discussion. 

## Why AGPL?
* Jan wants to build Nitro as a long-term ecosystem, and will drive Nitro's development for next few years
* We want strong copyleft culture, where downstream consumers improvements are fed back into the ecosystem
* We will be investing a lot of effort into integrating TensorRT-LLM and other libraries, and want potential Jan competitors to compete with us on an even playing field
* Longer-term, we can consider a dual-licensing scheme for Nitro, to waive AGPL with a commercial license (ala MySQL)

## Who else uses AGPLv3?
* MongoDB
* Minio
* Increasing numbers of the r/selfhosted stack uses AGPL, and we maintain license compatibility

cc @tikikun @vuonghoainam @0xSage

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,21 +1,660 @@
-MIT License
+# GNU AFFERO GENERAL PUBLIC LICENSE
 
-Copyright (c) 2023 Jan
+Version 3, 19 November 2007
 
-Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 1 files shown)


tikikun (approved):
LGTM

PR #142, opened 2023-11-15
Add nitro installation instruction

hiento09:
Add nitro installation instruction to docusaurus


hiento09:
Close since @hahuyhoang411 will include this docs in another PR

PR #154, opened 2023-11-21
fix navigating + change api path to api-ref

hahuyhoang411:
- [x]  Fix navigate bug
- [x] Change API path from `https://nitro.jan.ai/api` to `https://nitro.jan.ai/api-reference`
- [ ] Update new URL for fully compatible with OpenAI


tikikun (approved):
LGTM

PR #155, opened 2023-11-21
chore: enable-mermaid

hieu-jan:
PR Title: Enable mermaid
PR Description:
In this pull request, I modified `docusaurus.config.js` for config theme-mermaid. 
As our code base is configured under "@docusaurus/core": "^2.4.3", thus, every libraries `@docusaurus/...` needs to set the same version with core. Thus, I set to 2.4.3 and also add additional "@docusaurus/plugin-content-docs": "2.4.3" for running with mermaid.

Image of testing mermaid
image


tikikun (approved):
LGTM

PR #171, opened 2023-11-22
Add tar gz to nitro release and add docs publish instruction to README.md docs

hiento09:
Add tar gz to nitro release and add docs publish instruction to README.md docs


tikikun (approved):
LGTM

PR #177, opened 2023-11-22
feat(example): Add openai integration with nitro

hiro-v:
- [x] OpenAI-node integration with Nitro
- [x] OpenAI-python integration with Nitro


tikikun (approved):
LGTM

PR #188, opened 2023-11-24
Update views

hahuyhoang411:
- [x] Update codeblock views
- [x] Update FAQs views


hahuyhoang411:
If you have any magic trick to improve a dropdown, please do it. Thanks

PR #225, opened 2023-12-01
chore: Speedup windows build

vuonghoainam:
With --parallel in cmake based on drogon ci: https://github.com/drogonframework/drogon/blob/master/.github/workflows/cmake.yml#L61


vuonghoainam:
it's not correct to build like this, will revisit

PR #230, opened 2023-12-03
bug: Fix caching compatibility issues regarding system prompt

tikikun:
- This issue also seemed not addressed on the upstream repo, so i need to fix it on nitro level


tikikun:
Will provide significance boost for CPU only inference

PR #265, opened 2023-12-13
Add build target windows-oneAPI-build

### LINKED ISSUE #911: feat: optimization for oneAPI on windows
**Problem**
Currently the windows jan does not support oneAPI which will not provide the best performance for intel hardware

Alan successfully built Nitro with OneAPI using this docker environment:
```
# Use Debian stable slim as the base image
FROM intel/oneapi-basekit:2024.0.0-devel-ubuntu22.04

# Set working directory
WORKDIR /app

# Install required packages
RUN apt-get update && \
    apt-get install -y git cmake numactl uuid-dev && \
    git clone --recurse https://github.com/janhq/nitro && \
    cd nitro && \
    ./install_deps.sh && \
    mkdir build && \
    cd build && \
    cmake .. -DLLAMA_BLAS=ON -DLLAMA_BLAS_VENDOR=Intel10_64lp -DCMAKE_C_COMPILER=icx -DCMAKE_CXX_COMPILER=icpx && \
    cmake --build . --config Release -j $(nproc) && \
    apt-get remove --purge -y git cmake && \
    apt-get autoremove -y && \
    apt-get clean && \
    rm -rf /var/lib/apt/lists/*

# Expose the port
EXPOSE 3928

# Change the permissions of the nitro binary to make it executable
RUN chmod +x /app/nitro/build/nitro

# Set the command to run the nitro binary with numactl limiting to cores 0-7
ENTRYPOINT ["numactl", "--physcpubind=0-7", "/app/nitro/build/nitro"]
CMD ["1", "0.0.0.0", "3928"]
```
However the result binary does not work with system with without OneAPI runtime, tested on Debian base image. 
We need to do find a way to deliver this Nitro binary, with focus on Windows first, and later extends to other OS.

**Success Criteria**
Figure out a way to include OneAPI to windows. Either by statically compile OneAPI into Nitro binary, or delivering minimal OneAPI run-time with Nitro installation. 

Alan added comment: We need to have the nitro binary independence (no need for oneAPI runtime) or a good way to install oneAPI runtime to customer machine.

linhtran174:
Add OneAPI build for nitro. The performance gain is marginal and not even consistent. 
OneAPI doesn't seem to utilize all the cores though, it's using 1 CPU only, while my system has 2.  Same behavior with normal BLAS build.

MKL3

Build environment will need:
- Vcpkg: https://vcpkg.io/en/getting-started.html
--: set %PKG_CONFIG_PATH% 
- OneAPI: https://www.intel.com/content/www/us/en/docs/oneapi-base-toolkit/get-started-guide-windows/2024-0/overview.html
-- %ONEAPI_PATH% : path to `Intel/oneAPI` folder
-- %BLAS_INCLUDE_DIRS% : This is needed as FindBLAS from cmake have a bug getting the correct folder path. Its path is at /oneAPI/mkl/latest/include

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -386,6 +386,7 @@ jobs:
           cmake --build ./build_deps/nitro_deps --config Release
           mkdir -p build
           cd build
+          %ONEAPI_PATH%/setvars.bat
           cmake .. -DLLAMA_BLAS=ON -DLLAMA_BLAS_VENDOR=Intel10_64lp -DCMAKE_C_COMPILER=icx-cc -DCMAKE_CXX_COMPILER=icpx -DLLAMA_NATIVE=OFF -DLLAMA_BUILD_SERVER=ON -DBUILD_SHARED_LIBS=ON -DCMAKE_TOOLCHAIN_FILE=%PKG_CONFIG_PATH% -DBLAS_INCLUDE_DIRS=%BLAS_INCLUDE_DIRS%
           cmake --build . --config Release -j "%NUMBER_OF_PROCESSORS%"
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -351,6 +351,78 @@ jobs:
           asset_path: ./nitro.tar.gz
           asset_name: nitro-${{ needs.create-draft-release.outputs.version }}-win-amd64.tar.gz
           asset_content_type: application/gzip
+  
+  windows-oneAPI-build:
+    runs-on: windows-latest
+    needs: create-draft-release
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -351,6 +351,78 @@ jobs:
...
+          submodules: recursive
+
+      - name: Setup VSWhere.exe
+        uses: warrenbuckley/Setup-VSWhere@v1
+        with:
+          version: latest
+          silent: true
+        env:
+          ACTIONS_ALLOW_UNSECURE_COMMANDS: true
+
```
vuonghoainam, line 377: I don't see the step to install openapi toolkit for windows @linhtran174 ?

vuonghoainam:
The follow up issue seems closed

PR #276, opened 2023-12-16
feat: Add `/v1/audio/transcriptions` with llama.cpp

vuonghoainam:
- API: https://platform.openai.com/docs/api-reference/audio/createTranscription
- Whisper.cpp: https://github.com/ggerganov/whisper.cpp.git


tikikun (approved):
LGTM

PR #311, opened 2024-01-03
310 bug cannot unload model on nitro 022

tikikun:


tikikun:
- What this PR will fix:
+ Fix the wrong way to include clip (in previous version need manually now it's properly specified in cmakelist of llava)
+ Fix the CV inside new llama cpp update, we can revisit this later on for a bit of extra performance, right now it's better just do busy waiting using while loop

Issue #175, opened 2023-11-22
Discussion: Nitro to abstract Model Operations?


dan-jan:
## Objective

- I would like to float a thought balloon for Nitro to expand its feature set, and abstract model loading/unloading ops from the user
- The rise of multi-modal AI makes loading/unloading, VRAM and RAM management important
- This represents an opportunity for us to grow Nitro's value add layer, vs. just being a "server.cpp"

### Reasoning

- Developers new to local AI are used to OpenAI, where they don't have to think about model loading/unloading
   - Devs will just call `chat/completions` with different model names
   - Devs will just call `chat/completions` with images, sound, files etc
- As we progress towards multi-modal AI on resource-constrained devices, "model ops" is a concern
   - Hear, speak, think all require different models (even if we move to LlaVa style models)
   - Memory management, especially on constrained devices, is a field ripe for optimization
- Jan will primarily be a VSCode-tier product (i.e. application layer)
    - Nitro & Jan should have clean abstractions driven by architectural decisions
    - I see Jan's roadmap as being very driven by Assistants, while leaving local model management to Nitro
    - Imho, this functionality should live with Nitro

### Nitro be "model aware"

- A Nitro server would be "aware" of what models it has access to
- Naturally, this would be via a `/models` folder (aligning with Jan has significant business benefits for us)
- This would allow users to pass in `model` params in `chat/completions`, similar to OpenAI

### Nitro to abstract model loading/unloading from user

- Devs would just send `chat/completions` requests to Nitro
- Nitro will handle loading/unloading (we just assume requests can always be queued in-memory)
- Nitro -may- send "statuses" via SSE to requester, to indicate model loading time lag (Dan's question: is this a good idea?)

### Nitro to have a `/models` endpoint

We would refactor current Nitro models endpoints: 

| Status Quo | Recommended Change | Function | 
|-|-|-|
| /modelstatus | GET v1/models | Return state of models in Nitro |
| /loadmodel | POST v1/models/load { ... params } | Loads model |
| /unloadmodel | POST v1/models/unload { ...params } | Unloads model |

### Nitro to be aware of available RAM/VRAM

- Debatable: Nitro should be aware of available RAM/VRAM
- Nitro should figure out when to unload models (a lot of good techniques from RAM/Memory principles)

### Open Questions

Q: What happens if a user wants to load a model with specific params?
A: Nitro still retains a `POST models/load` endpoint for this purpose, and subsequent calls to Model will have those params (unless unloaded)
[...]

PR #318, opened 2024-01-05
[WIP] Nitro model management

hiro-v:
Feature for https://github.com/janhq/nitro/issues/175

- [x] Load multiple models
- [ ] Add GET `models` to return models list
- [ ] CUDA support for multiple model request at the same time (CCU 1 for each model)
- [x] Metal support for multiple model request at the same time (CCU 1 for each model)
- [x] CPU support for multiple model request at the same time (CCU 1 for each model)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -57,8 +57,13 @@ add_executable(${PROJECT_NAME} main.cc)
 #
 # and comment out the following lines
 find_package(Drogon CONFIG REQUIRED)
+
+# Add boost
+find_package(Boost COMPONENTS filesystem REQUIRED)
+target_include_directories(${PROJECT_NAME} PRIVATE ${Boost_INCLUDE_DIRS})
+
 target_link_libraries(${PROJECT_NAME} PRIVATE Drogon::Drogon common llama llava
-                                              ${CMAKE_THREAD_LIBS_INIT})
+                                              ${CMAKE_THREAD_LIBS_INIT} ${Boost_LIBRARIES})
 
 # ##############################################################################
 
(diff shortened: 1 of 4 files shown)


hiro-v:
CleanShot 2024-01-08 at 09 43 30
CleanShot 2024-01-08 at 09 43 37
CleanShot 2024-01-08 at 09 43 51
CleanShot 2024-01-08 at 09 44 02
CleanShot 2024-01-08 at 09 44 17
CleanShot 2024-01-08 at 09 44 25

hiro-v:
Test result for running 2 concurrent request to 2 llama.cpp on the same process given the change: It still crashes.

```
GGML_ASSERT: /home/hiro/jan/nitro/llama.cpp/ggml-cuda.cu:6742: ptr == (void *) (g_cuda_pool_addr[device] + g_cuda_pool_used[device])
```

The line: https://github.com/ggerganov/llama.cpp/blob/1bf681f90ef4cf37b36e6d604d3e30fc57eda650/ggml-cuda.cu#L6742

Trace to variable `g_cuda_pool_addr` and `g_cuda_pool_used`
https://github.com/ggerganov/llama.cpp/blob/1bf681f90ef4cf37b36e6d604d3e30fc57eda650/ggml-cuda.cu#L6664
```
static CUdeviceptr g_cuda_pool_addr[GGML_CUDA_MAX_DEVICES] = {0};
static size_t g_cuda_pool_used[GGML_CUDA_MAX_DEVICES] = {0};
```
These are static global variable and cannot be shared across thread in the same application.
This global pool thing is everywhere across the llama cpp cuda code base. So refactoring would be hard + required extensive knowledge in llama cpp.

hiro-v:
I also added
```
"n_parallel": 2,
"cont_batching": true
```
But it does not help on CUDA. 
However on CPU, Mac Metal it works flawlessly without the need to add extra variables

PR #320, opened 2024-01-07
Update README.md

eltociear:
Github -> GitHub


tikikun:
Thanks for reminding i will directly change on main branch

Issue #347, opened 2024-01-16
feat: Support GBNF Grammars


LouisIV:
**Problem**
Currently there is no support for GBNF grammars in the API. If there is its poorly documented.

**Success Criteria**
Llama.cpp supports GBNF grammars which is really useful when trying to emulate OpenAI function calling (this is actually the approach that LocalAI takes to do this). I would love to see that support exposed in some manner.

**Additional context**
n/a

tikikun:
it's already support in https://github.com/janhq/nitro/pull/348

PR #348, opened 2024-01-16
add support for gbnf grammar

tohrnii:
This PR adds support for gbnf grammars from upstream llama.cpp.
Addresses #347


tikikun (approved):
LGTM

Issue #295, opened 2023-12-21
feat: Cortex supports Function Calling


ChristianWeyer:
## Goal

- llama3.1 should support Tool use in llama.cpp
- https://github.com/janhq/models/issues/16

## Questions

- Is our implementation of Function Calling consistent across all models?
    - e.g. do we need `model.yaml` to have any specific template?

## Original post

**Problem**
AFAICS, the current implementation does not have OpenAI Function Calling support. This would be a fantastic, powerful, and much needed feature.

**Success Criteria**
Any OAI client can be used with Nitro, even (and especially) those that use OAI Function Calling.

Reference:
https://platform.openai.com/docs/guides/function-calling
https://platform.openai.com/docs/api-reference/chat/create#chat-create-tools

## Linked Issues
- [Jan #3785 Support response_format and structured JSON responses](https://github.com/janhq/jan/issues/3785)

## Bugs
- [ ] #1389

nidhoggr-nil:
Would be great to have this for the tensorRT backend as well, though I do not know if they support it.

dan-menlo:
I am scoping this to llama3.1 function calling, as per discussion here https://github.com/janhq/cortex.cpp/issues/1151#issuecomment-2339558655

dan-menlo:
@nguyenhoangthuan99 I realize I did not scope this story well for Sprint 21: 

Edit: Found the comment https://github.com/janhq/models/issues/16

User story: Developer should be able to use Cortex with llama3.2, to do Function Calling (with an API similar to OpenAI)

Can you help me with the following:
- Add links to the existing Github gist you had for llama3.1 Function Calling support
- Help me think through how we should scope this - is there a way we can "mimic" the OpenAI API structure (e.g. pass in the function call into context window)

dan-menlo:
@nguyenhoangthuan99 Found your comment on Llama3.1 function calling, and pasting it here:

https://github.com/janhq/models/issues/16

Here is python script that run function calling, GitHub block `.py` file, so I cannot upload it.

```
import requests, json

ENDPOINT = "https://litellm.jan.ai/v1/chat/completions" # "http://localhost:3928/v1/chat/completions" #
MODEL = "alan-gift" # "meta-llama3.1-8b-instruct" #

grammar = """
root   ::= object
value  ::= object | array | string | number | ("true" | "false" | "null") ws

object ::=
  "{" ws (
            string ":" ws value
    ("," ws string ":" ws value)*
  )? "}" ws

[...]

PR #354, opened 2024-01-17
[WIP] add support for function calling

tohrnii:
This PR adds support for function calling / tool use. 
Addresses #295

Request:

```bash
curl http://localhost:3928/v1/chat/completions \
-H "Content-Type: application/json" \
  -d '{
    "messages": [
      {
        "role": "system",
        "content": "You are a helpful assistant with access to some tools. The list of tools and their uses are provided to you. You should respond with the tool to use and appropriate parameters. For example, if there is a calculator tool available and the user asks to do some arithmetic calculation, you can call the calculator() tool.
For eg:
input: What is 42*42?
output: {\"thought\": \"User wants to calculate multiplication of two numbers so I should use the calculator tool\", \"tool\": \"caculator(42*42)\"}
Here is the list of tools provided to you."
      },
      {
        "role": "user",
        "content": "What is the weather like in SF?"
      },
    ], "grammar_file": "/path/to/grammar_file",
    "tools": "[{
      \"type\": \"function\",
      \"function\": {
        \"name\": \"get_current_weather\",
        \"description\": \"Get the current weather in a given location\",
        \"parameters\": {
          \"type\": \"object\",
          \"properties\": {
            \"location\": {
              \"type\": \"string\",
              \"description\": \"The city and state, e.g. San Francisco, CA\"
            },
            \"unit\": {
              \"type\": \"string\",
              \"enum\": [\"celsius\", \"fahrenheit\"]
            }
          },
          \"required\": [\"location\"]
        }
      }
    }]",
  "tool_choice": "auto"
  }'
```

Response:
```
{\"thought\": \"User wants to know the weather in San Francisco, so I should use the get_current_weather tool\", \"tool\": \"get_current_weather(location='San Francisco', unit='celsius')\"}
```

### DIFF
--- a/controllers/llamaCPP.cc
+++ b/controllers/llamaCPP.cc
@@ -193,6 +193,8 @@ void llamaCPP::chatCompletion(
     data["frequency_penalty"] =
         (*jsonBody).get("frequency_penalty", 0).asFloat();
     data["presence_penalty"] = (*jsonBody).get("presence_penalty", 0).asFloat();
+    std::string tools = (*jsonBody).get("tools", "").asString();
+    LOG_INFO << tools;
     const Json::Value &messages = (*jsonBody)["messages"];
     std::string grammar_file = (*jsonBody).get("grammar_file", "").asString();
     std::ifstream file(grammar_file);
@@ -204,7 +206,7 @@ void llamaCPP::chatCompletion(
       data["grammar"] = grammarBuf.str();
     }
     if (!llama.multimodal) {
-
+      bool systemPromptProcessed = false;
       for (const auto &message : messages) {
         std::string input_role = message["role"].asString();
         std::string role;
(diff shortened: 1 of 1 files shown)


tohrnii:
I guess the big question here is what format should the tools input be in if we also want to use llms that are not finetuned for tool use. llama-cpp-python seems to format the tools into something like:

```
// Supported function definitions that should be called when necessary.
namespace functions {

  // Get the current weather in a given location
  type get_current_weather = (_: {
    // The city and state, e.g. San Francisco, CA
    location: string,
    // Optional parameter indicating temperature unit
    unit?: "celsius" | "fahrenheit",
  }) => any;

} // namespace functions
```
(I could be wrong, this is just from reading the code)

There is an argument to be made that maybe it should resemble a python function instead as most llms should have seen a lot of python code and should know how to invoke python functions better? And then leave it to the consumer to do a validation pass before doing another inference pass. Is there a better way to accomplish this?

In my limited testing with mistral and llama2-chat, I have got decent results with just passing the json as it is unless the user intentionally sends malformed input.

louis-jan:
It's pretty outdated, any update on this @tohrnii? Could you please share an example or so and reopen after rebased?

louis-jan:
KIV @vansangpfiev

PR #367, opened 2024-01-22
[janhq/jan#1635] feat(WIP): decouple nitro engine into a library

InNoobWeTrust:
## Current progress
- The library can be built, but with local dependency on janhq/jan as submodule
- Next step is update janhq/core to npmjs and use the resolved package there instead of submodule
- TODO: update GitHub-action workflow to build nitro-node for different OSes and add the tarballs of npm packages to release
- TODO: from janhq/jan build step for extensions, simply download nitro-node from github release

## The plan:
Move the nitro extension to nitro repo
Build the extension and put tarballs into release together with nitro
From Jan, will need to add some kind of version.txt for downloading correct version of nitro-node and download it to put into ./electron/pre-install/

## Issue:
Currently I'm cloning https://github.com/janhq/jan as a submodule in order to build the janhq/core package. The version on npm seem to be outdated and the imports do not work although the version is displayed to be the same (https://www.npmjs.com/package/@janhq/core) => can you update it on npmjs or the team prefer to coupled the versions tightly with github repo?


InNoobWeTrust:
Test build on different yarn versions (platform: MacOS M1):
- v1.22.21: build success
- v2.4.3: build fail with error `➤ YN0066: │ typescript@patch:typescript@npm%3A5.3.3#builtin<compat/typescript>::version=5.3.3&hash=8133ad: Cannot apply hunk #1 (set enableInlineHunks for details)`
- v3.7.0: build success

How to reproduce:
```sh
npm i -g corepack
corepack enable  --install-directory <some_dir_in_PATH>
corepack install -g yarn@<1,2,3>
```

nitro-node/src/index.ts
```diff
@@ -0,0 +1,277 @@
...
+  fs,
+  Model,
+  joinPath,
+  InferenceExtension,
+  log,
+  InferenceEngine,
+  MessageEvent,
+  ModelEvent,
+  InferenceEvent,
+} from "@janhq/core";
```
louis-jan, line 27: I see that you moved the entire nitro extension source code into this package. This would couple the Jan app business logic, which is not the idea of the nitro-node, as a stand-alone module.

1. How do you define these logics? For example, how do they work outside of the Jan app (other node apps)?
2. Is this a browser module or a Node.js module?

nitro-node/src/@types/global.d.ts
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * The parameters for the initModel function.
+ * @property settings - The settings for the machine learning model.
+ * @property settings.ctx_len - The context length.
+ * @property settings.ngl - The number of generated tokens.
+ * @property settings.cont_batching - Whether to use continuous batching.
+ * @property settings.embedding - Whether to use embedding.
+ */
+interface EngineSettings {
```
louis-jan, line 13: Decouple Jan business logic from nitro-node.

nitro-node/src/module.ts
louis-jan, line 1: The file is redundant. Node module already defined.

nitro-node/src/index.ts
InNoobWeTrust: Noted, after the build workflow for both Jan and nitro-node can be run independently as of now I'm beginning to refactor the source code inside nitro-node. Should take some time before the PR is ready...🧐

nitro-node/src/nvidia.ts
```diff
@@ -0,0 +1,201 @@
...
+    version: "",
+  },
+  gpus: [],
+  gpu_highest_vram: "",
+};
+
+/**
+ * Path to the settings file
+ **/
+export const NVIDIA_INFO_FILE = path.join(
```
louis-jan, line 27: Remove Application business logic

nitro-node/test/model.json
```diff
@@ -0,0 +1,28 @@
+{
```
louis-jan, line 1: What is the purpose of this mock file? I'm seeing that the test settings will need only 2 values: model's URL & template.

nitro-node/src/index.ts
```diff
@@ -0,0 +1,348 @@
...
+  killSubprocess();
+}
+
+export default {
+  setLogger,
+  runModel,
+  stopModel,
+  killSubprocess,
+  dispose,
+  updateNvidiaInfo,
```
louis-jan, line 346: The module should not depend on the app's business logic. Let's assume that the app's setting folder directory might be different between apps, which could confuse users. The module should be stateless or sandboxed & the folder might not exist.

nitro-node/src/index.ts
```diff
@@ -0,0 +1,348 @@
...
+  // clean other registered resources here
+  killSubprocess();
+}
+
+export default {
+  setLogger,
+  runModel,
+  stopModel,
+  killSubprocess,
+  dispose,
```
louis-jan, line 345: From module level, they are the same function but different endpoints. Dispose is the extension's logic as well. Please remove and leave it a clearer name.

.github/scripts/auto-sign.sh
louis-jan, line 1: Code signing should be done from application level not from module level.

nitro-node/test/model.json
InNoobWeTrust: This json file is copied from Jan downloaded model directories for a quick test on the implementation, will remove unnecessary config then.

.gitignore
```diff
@@ -86,6 +86,7 @@ CMakeFiles
 CMakeScripts
 Testing
 Makefile
+!nitro-node/Makefile
```
louis-jan, line 89: Why ignore then unignore?

nitro-node/.gitignore
```diff
@@ -0,0 +1,23 @@
...
+
+*.log
+
+# Nitro binary files
+bin/*/nitro
+bin/*/*.metal
+bin/*/*.exe
+bin/*/*.dll
+bin/*/*.exp
+bin/*/*.lib
```
louis-jan, line 23: What files should we keep?

nitro-node/.gitignore
```diff
@@ -0,0 +1,23 @@
+.vscode
+.env
+
+# Jan inference
+error.log
```
louis-jan, line 5: *.log already covered this

.gitignore
InNoobWeTrust: Makefile that is generated by Cmake for nitro (cpp) is ignored from root (can alsway specify the output build directory when invoking Cmake). But I need to put some utilities for CI workflow in `nitro-node/Makefile`. For now, just use Makefile for CI tasks definition for `@janhq/nitro-node` until I can find a better replacement.

nitro-node/.gitignore
InNoobWeTrust: We don't need them in sourcecode, as there is already a `postinstall` hook in `package.json` that will download all the binaries into the correct place if `@janhq/nitro-node` is installed as a dependency. I tested with both npm and yarn@1 and that works.

nitro-node/.gitignore
InNoobWeTrust: Noted, will remove.

nitro-node/package.json
```diff
@@ -0,0 +1,62 @@
...
+  "node": "dist/index.cjs.js",
+  "author": "Jan <service@jan.ai>",
+  "license": "AGPL-3.0",
+  "scripts": {
+    "test": "jest --verbose --detectOpenHandles",
+    "pretest": "node download-nitro.js",
+    "build": "tsc --module commonjs && rollup -c rollup.config.ts",
+    "downloadnitro": "node download-nitro.js",
+    "build:publish": "npm pack",
+    "postinstall": "node postinstall.js"
```
louis-jan, line 15: What would happen if there were a module that depend on this module? e.g. module wrapper such as nitro-extension. 
1. Will post-install work with "npm install" command?
2. Will post-install step runs as soon as user install wrapper module?

nitro-node/.gitignore
louis-jan: So pattern ignore should work? e.g. bin or bin*/*.*

nitro-node/package.json
InNoobWeTrust: I packed `nitro-node` as a tarball and then install from that and it works. But will need to test for some more cases. Let me setup CI later to verify this, if not then I'll expose the function to let wrapper download the binaries.

nitro-node/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "@types/jest": "^29.5.11",
+    "@types/node": "^20.11.4",
+    "@types/os-utils": "^0.0.4",
+    "@types/tcp-port-used": "^1.0.4",
+    "jest": "^29.7.0",
+    "rimraf": "^3.0.2",
+    "rollup": "^2.38.5",
+    "rollup-plugin-define": "^1.0.1",
+    "rollup-plugin-sourcemaps": "^0.6.3",
+    "rollup-plugin-typescript2": "^0.36.0",
```
louis-jan, line 36: I don't see these dependencies are used anywhere in the module now.

nitro-node/package.json
InNoobWeTrust: Please see the fourth commit here: https://github.com/janhq/jan/pull/1712/commits
I installed from git subdirectory using gitpkg and the binaries are downloaded in `node_modules/@janhq/nitro-node` although in the source there is no `bin`.

nitro-node/package.json
InNoobWeTrust: Noted, will check and remove unused ones.

nitro-node/package.json
louis-jan: > I packed `nitro-node` as a tarball and then install from that and it works. But will need to test for some more cases. Let me setup CI later to verify this, if not then I'll expose the function to let wrapper download the binaries.

I mean, you packaged as a tarball on MacOS computer, there would be MacOS bin files, but when you put the Tarball in Windows, that would break.

louis-jan: Please test with both NPM & Yarn, scenario:
User might just add nitro-node into their package dependencies, it should work without any additional steps (only npm / yarn install)

nitro-node/.gitignore
InNoobWeTrust: I think just need to ignore the whole `bin` directory. Let me update the ignore file.

nitro-node/package.json
InNoobWeTrust: Ah, I understand your point, the tarball should be generic so there is no `bin` in it also, as it is not defined in the `files` properties of `package.json`. Let me continue the verification of `postinstall` hook when I make the github action workflow for `nitro-node` then.

InNoobWeTrust:
Tests passed on Ubuntu (no CUDA) and Mac-Intel: https://github.com/InNoobWeTrust/nitro/actions/runs/7652932312
- start/stop nitro process
- run `chatCompletion` with stream option

InNoobWeTrust:
Test cases for installing on multiple platforms:
- Test install as direct dependency => pass for both npm and yarn
- Test install as indirect dependency => success on windows, fail npm for linux and macos (wrong path is resolved if install from local directory with its own dependency)

Test results with both cases: https://github.com/InNoobWeTrust/nitro/actions/runs/7660930982
Test results when disable the indirect dependency test case: https://github.com/InNoobWeTrust/nitro/actions/runs/7661001110

As currently the most used scenario is for Jan extension to use a direct dependency so it will be fine.

nitro-node/package.json
InNoobWeTrust: Please see [this comment](https://github.com/janhq/nitro/pull/367#issuecomment-1911079924) for the test results regarding this discussion.

nitro-node/dist/download-nitro.cjs.js
louis-jan, line 1: Ignore /dist

InNoobWeTrust:
Workflow for testing installation now use tarball built in linux runner for installation test on all the supported platforms:
https://github.com/InNoobWeTrust/nitro/actions/runs/7665077994

InNoobWeTrust:
> > Workflow for testing installation now use tarball built in linux runner for installation test on all the supported platforms: https://github.com/InNoobWeTrust/nitro/actions/runs/7665077994
> 
> The tarball does not include nitro execution files, so that mean users have to download it manually at runtime? Here is the one you might want to take a look: https://www.npmjs.com/package/@mapbox/node-pre-gyp

Runtime function `runModel` also downloads nitro binaries if they are not found. The binaries is downloaded given following scenarios:
- `@janhq/nitro-node` is added as dependency and the lifecycle hook `postinstall` is called by package manager (only verified that the lifecycle is compatible with npm and yarn at the moment)
- During runtime, when calling `runModel`, the expected binary dir is checked, if the binaries are not downloaded then they will be downloaded automatically. I have exposed a function to let caller explicitly set their desired binary dir also, so it will be flexible on where users expect the binaries to be in system path.

InNoobWeTrust:
`@janhq/nitro-node` is proved to work when integrating with Jan. I'm changing this PR to ready-for-review.

nitro-node/src/scripts/download-nitro.ts
```diff
@@ -0,0 +1,219 @@
...
+
+// Define nitro version to download in env variable ("latest" or tag "v1.2.3")
+const NITRO_VERSION = process.env.NITRO_VERSION || "latest";
+// The platform OS to download nitro for
+const PLATFORM = process.env.npm_config_platform || process.platform;
+// The platform architecture
+//const ARCH = process.env.npm_config_arch || process.arch;
+
+const releaseUrlPrefixVariants = {
+  latest: "https://api.github.com/repos/janhq/nitro/releases/",
```
louis-jan, line 14: Please leave these configs in rollup.config.js. Some forks may find it difficult to locate and replace here and there.

nitro-node/src/nitro.ts
```diff
@@ -0,0 +1,511 @@
...
+}
+
+/**
+ * Initializes a Nitro subprocess to load a machine learning model.
+ * @param modelFullPath - The absolute full path to model directory.
+ * @param promptTemplate - The template to use for generating prompts.
+ * @returns A Promise that resolves when the model is loaded successfully, or rejects with an error message if the model is not found or fails to load.
+ */
+export async function runModel({
+  modelFullPath,
```
louis-jan, line 168: Can rename it to modelPath

nitro-node/src/nitro.ts
```diff
@@ -0,0 +1,511 @@
...
+ * @returns A Promise that resolves when the subprocess is terminated successfully, or rejects with an error message if the subprocess fails to terminate.
+ */
+export function stopModel(): Promise<NitroModelOperationResponse> {
+  return killSubprocess();
+}
+
+/**
+ * Read the magic bytes from a file and check if they match the provided magic bytes
+ */
+export async function checkMagicBytes(
```
louis-jan, line 141: Refactor to utils

nitro-node/src/nitro.ts
```diff
@@ -0,0 +1,511 @@
...
+      log(`[NITRO]::Error: ${data}`);
+    });
+
+    subprocess.on("close", (code: any) => {
+      log(`[NITRO]::Debug: Nitro exited with code: ${code}`);
+      subprocess = undefined;
+      reject(`child process exited with code ${code}`);
+    });
+
+    tcpPortUsed.waitUntilUsed(PORT, 300, 30000).then(() => {
```
louis-jan, line 481: Sorry for the confusion, I don't think spawning a nitro subprocess should take that long. Please help adjust.

nitro-node/src/nitro.ts
```diff
@@ -0,0 +1,511 @@
...
+const NITRO_HTTP_VALIDATE_MODEL_URL = `${NITRO_HTTP_SERVER_URL}/inferences/llamacpp/modelstatus`;
+// The URL for the Nitro subprocess to kill itself
+const NITRO_HTTP_KILL_URL = `${NITRO_HTTP_SERVER_URL}/processmanager/destroy`;
+// The URL for the Nitro subprocess to run chat completion
+const NITRO_HTTP_CHAT_URL = `${NITRO_HTTP_SERVER_URL}/inferences/llamacpp/chat_completion`;
+
+// The default config for using Nvidia GPU
+const NVIDIA_DEFAULT_CONFIG: NitroNvidiaConfig = {
+  notify: true,
+  run_mode: "cpu",
```
louis-jan, line 44: These are not related to nitro-node and GPU.

louis-jan: Can consider run_mode, but should not be overwritten. Can set default "gpu" when its not defined and CUDA check passed. Otherwise app can overwrite (nitro extension should cache it somewhere, e.g. settings.json before)

InNoobWeTrust: Oops, missed this detail. Will refactor the logic for this, this field should be an optional parameter to be specified at runtime, not in the runtime-persistent config passed around globally.

nitro-node/src/scripts/download-nitro.ts
```diff
@@ -0,0 +1,220 @@
...
+import crossFetch from "cross-fetch";
+
+// Define nitro version to download in env variable ("latest" or tag "v1.2.3")
+const NITRO_VERSION = process.env.NITRO_VERSION || "latest";
+// The platform OS to download nitro for
+const PLATFORM = process.env.npm_config_platform || process.platform;
+// The platform architecture
+//const ARCH = process.env.npm_config_arch || process.arch;
+
+const getReleaseInfo = async (taggedVersion: string): Promise<any> => {
```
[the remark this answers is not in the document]
InNoobWeTrust: I prefer to skip node-gyp at this stage. As I understand, the primary use-case of node-gyp is to build node native modules (eg. when we want to build nitro into *.so files and use node binding to control nitro directly instead of going through a subprocess wrapper).
Using node-gyp for just getting the binaries is somehow overkill, and supporting programmatic build on every supported platform is another story. At this point I'm not confident enough that I can support it without people raising issues (build env involves PATH, OS versions, available tooling, etc..). For now, I don't think stepping into it will make good use of the time, more issues will happen...

nitro-node/src/prompt.ts
```diff
@@ -0,0 +1,50 @@
+import { NitroPromptSetting } from "./types";
+
+/**
+ * Parse prompt template into agrs settings
+ * @param {string} promptTemplate Template as string
+ * @returns {(NitroPromptSetting | never)} parsed prompt setting
+ * @throws {Error} if cannot split promptTemplate
+ */
+export function promptTemplateConverter(
```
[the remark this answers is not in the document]
InNoobWeTrust: You're right, but for the sake of this PR, let's just create another issue for the template converter. Gonna check the flow of C++ later and adapt this in another PR.

nitro-node/src/nvidia.ts
```diff
@@ -0,0 +1,158 @@
...
+  }
+
+  nvidiaConfig["cuda"].exist = cudaExists;
+  nvidiaConfig["cuda"].version = cudaVersion;
+}
+
+/**
+ * Get GPU information
+ */
+async function updateGpuInfo(): Promise<void> {
```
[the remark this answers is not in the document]
InNoobWeTrust: I'm not sure about this, do you think most people will use the library with responsibility (exporting `CUDA_VISIBLE_DEVICES`) or prefer it fool-proof as default and let the library define sensible options for them? 🤔

nitro-node/src/execute.ts
```diff
@@ -0,0 +1,64 @@
...
+export interface NitroExecutableOptions {
+  executablePath: string;
+  cudaVisibleDevices: string;
+}
+
+/**
+ * Find which executable file to run based on the current platform.
+ * @returns The name of the executable file to run.
+ */
+export const executableNitroFile = (
```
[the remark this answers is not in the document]
InNoobWeTrust: - I don't quite understand this, nitro-node only helps to get nitro binaries available on the user's machine, no downloading of the model here...😅
- Ok, will go with this. But the logic will be not flexible in the early versions, need to discuss the naming schemes of nitro release archives (can we have only a single generic CUDA build without separating into different versions?)
- The scope of this PR is to first split the logic from Jan into a library here for re-use and easier to maintain as it will be tightened with nitro development. I'm not yet involved much in the architecture/build decision of nitro so for now I will just stick with the subprocess. Until nitro can be used as a system library then make a node binding later (will be much more work). Let's go with the iterative approach and create another issue for Node binding at the moment, I don't want the PR to become too big to review with unplanned things...😅
- I believe we can't handle all the unplanned things all at once or we will go nowhere. It might be the use case but not until it's confirmed to be worth our effort (user survey and confirmed to be valid first before doing). If things are not planned yet and not an agreement on the roadmap then I don't think we should spend much time over-engineering it. Just delivering features one step at a time...

For your suggestions, I'm considering and thinking about the best ways to adapt for now with minimal effort possible. But for things in the future, please share the roadmap link of the mentioned point and a defined plan so I can understand what are the steps to iteratively deliver it. I'm not a team member at the moment so I don't know well about your plans so pardon if I'm not understanding your points correctly...😅

nitro-node/src/nvidia.ts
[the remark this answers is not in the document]
InNoobWeTrust: Yes, I mean if users do not provide anything then the library will use defaults based on analysis. And for the one knowing and wanting to be specific, we let them specify it.
For now, I'll add support for env var to let users specify their preferred devices.

nitro-node/src/scripts/download-nitro.ts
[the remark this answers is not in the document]
InNoobWeTrust: The intention for now is simple, to ease the usage of nitro in Nodejs programs without too much boilerplate code repeated and bloated between projects. It's not about overhead but rather about practical usage, it's just a simple wrapper for now, with no intention to bring surplus value to what nitro already has but rather improve developer experience on making a client program that utilizes nitro. 😅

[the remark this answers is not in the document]
InNoobWeTrust: For building nitro during installation if no binary is available for the platform, let's move it to another issue as I believe it will be rather complex. Let's deliver what is useful for gaining more adoption of nitro from fellow developers then improve it later with more features...😅

nitro-node/src/utils/index.ts
```diff
@@ -0,0 +1,40 @@
+import fs from "node:fs";
+import osUtils from "os-utils";
+import { log } from "../logger";
+import { ResourcesInfo } from "../types";
+
+/**
+ * Get the system resources information
+ */
+export async function getResourcesInfo(): Promise<ResourcesInfo> {
```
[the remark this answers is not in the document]
InNoobWeTrust: This one is for getting the number of CPU cores on the machine to provide a default value if not specified when running nitro. It's also useful for ones who use nitro-node, I believe.

nitro-node/src/scripts/download-nitro.ts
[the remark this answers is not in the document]
InNoobWeTrust: FYI: I'm looking at a reference build script at the link below
https://github.com/withcatai/node-llama-cpp/blob/master/src/cli/commands/BuildCommand.ts

They also don't use node-gyp but rather have a build script themself. I also prefer it this way, to not rely on node-gyp if not necessary means no dependency on Python and other native libs required by node-gyp, then we can have a less demanding library => more usage from other projects => more visibility, and more contribution back to nitro and this library. 😉

louis-jan:
@InNoobWeTrust @hiro-v I would like to proceed and merge this pull request. Any further updates will be in separate PRs.

louis-jan (approved):
LGTM

louis-jan (approved):
LGTM. Let's resolve the pending ones in the next PRs.

PR #368, opened 2024-01-22
Add AVX and AVX2 flag to support old and new CPU gen

hiento09:
Add AVX and AVX2 flag to support old and new CPU gen


hiento09:
Close this PR since both flag LLAMA_AVX and LLAMA_AVX2 is enable by default

PR #380, opened 2024-01-26
add checking model loaded first in embedding call

wujjpp:
checking model loaded first in embedding call, otherwise nitro hang


tikikun (approved):
LGTM

PR #383, opened 2024-01-29
fix: CI error on mac silicon and windows

hiro-v:
- [x] Mac silicon whisper.cpp (coreML build) error
- [x] Windows amd64 llama.cpp e2e test
- [x] Windows amd64 with CUDA llama.cpp e2e test - This cannot be run on CI because Nitro CI uses nvidia docker and does not have mounted physical device


tikikun (approved):
LGTM

PR #405, opened 2024-02-01
feat: Add handler for drogon to try binding port

hiro-v:
- [x] Test on Mac
- [x] Test on Linux
- [ ] Test on Windows


tikikun:
Nitro should not retry in order to introduce even more issues when debugging, for this step specifically it should fail easy by design

PR #412, opened 2024-02-02
feat: add queue for embedding

tikikun:


tikikun:
Please help me test if it can fix RAG issue, if ok please review i will merge @hiro-v

PR #424, opened 2024-02-06
ci: Add github workflow for nitro node

InNoobWeTrust:
This is a successor of janhq/nitro#367
- Fix bug not being able to spawn nitro process on Windows
- Add CI workflows for nitro-node


nitro-node/src/nitro.ts
```diff
@@ -214,23 +217,17 @@ async function runNitroAndLoadModel(
      * The tested threshold is 500ms
      **/
     if (process.platform === "win32") {
-      return await new Promise((resolve) => setTimeout(() => resolve({}), 500));
-    }
-    const spawnResult = await spawnNitroProcess(runMode);
-    if (spawnResult.error) {
-      return spawnResult;
+      await new Promise((resolve) => setTimeout(resolve, 500));
```
InNoobWeTrust, line 220: Hello @louis-jan, can you help review this PR? There is a serious bug at this line that I overlooked. The extra `return` makes the library unusable on Windows as the code will return early before spawning nitro process (my bad). 😫

louis-jan: Sure @InNoobWeTrust. Thank you

.github/workflows/build.yml
```diff
@@ -2,7 +2,7 @@ name: CI
 
 on:
   schedule:
-    - cron: "0 20 * * *" # At 8 PM UTC, which is 3 AM UTC+7
```
louis-jan, line 5: It's not relevant to the PR right?

InNoobWeTrust: yes, reverted the unrelated change. 😉

louis-jan (approved):
LGTM

louis-jan:
We are working on Cortex NPM and CLI so we could close this now.

InNoobWeTrust:
> We are working on Cortex NPM and CLI so we could close this now.

Great, I'm waiting for the official NPM package then. 😆

Issue #430, opened 2024-02-06
bug: model causing an unknown long wait for a response (and other issues)


Glashkoff:
**Describe the bug**
In our Discord, people mention the low speed of the second response or an empty response. I came across a model that causes all these problems in 0.4.6:
1) Attempting to regenerate the response may hangs on the progress bar "Generating response" (thread and log of this issue in the attached zip archive)
2) During the conversation, the speed decreases with each new response
3) Sometimes instead of a response - an empty message
4) Sometimes the response repeats infinitely without stop

This model: [stable-code-3b.Q8_0.gguf](https://huggingface.co/brittlewis12/stable-code-3b-GGUF/blob/main/stable-code-3b.Q8_0.gguf)
model.json is default for 0.4.6
It is clear that there is some incompatibility with the default Jan settings here, but it would be nice to somehow handle this or inform the user that something is wrong. 
I hope this example will help diagnose and fix or somehow handle the described problems.

**Steps to reproduce**
Download my thread.json and try to generate a response with the model from the link above in this thread. The result is unpredictable, but in 9 out of 10 cases - it's an irrelevant answer to the query in the form of one or several paragraphs of text consisting of words or code.

**Expected behavior**
For this model, special prompt_template and probably retrieval_template are needed. Therefore, under the current conditions, an inadequate response is expected in terms of meaning. However, it should be at least some with stable speed and without freezing the progress bar for generating the answer.

**Environment details**
Operating System: Windows 11 Pro x64 23H2 (build 22631.3085)
Jan Version: 0.4.6
Processor: AMD Ryzen 5 5600
RAM: 64GB
Any additional relevant hardware specifics: RTX 3060 12GB, driver 551.23, rebar on, Cuda system fallback policy disabled

**Logs**
[jan_1707226938.zip](https://github.com/janhq/jan/files/14181251/jan_1707226938.zip)

0xSage:
It's weird that users are seeing more degradation when they go from `online to offline` though. Worth investigating.

Brog33:
i am experiencing this same issue from time to time. tell me if you need some help testing.

tikikun:
alread fixed

PR #431, opened 2024-02-11
fix: race condition issue

louis-jan:
## Description
chunked_content_provider lambda race condition issue when: 
- T1: R1 start -> create run loop lambda and request completion. 
streaming = true | serial queue busy = true.
- T2: R1 done ->  
streaming = false | serial queue busy = false | stop = true (cancel completion request)
- T3: R1 run loop still run -> 
recreate request completion for a `completed request`? -> `zombie request` (sus memory leak?).
- T2: R1 completion request cancelled, R2 come -> llama.next_result error -> `jump in a loop`, R1 back to run loop and create new request completion -> 
- T1.5: lost connection -> R1 cancel request in progress -> R2 come & create completion -> single_queue_is_busy true, next_result error -> `jump in a loop` R3 come, stuck.

1. Its better to not create completion request within lambda, should create before triggering (serially). Apply a serial queue to handle this. Can be easily replaced by Concurrent Queue with workers number = n_parallel later.

This is to make it safer by not triggering 2 lambda functions at the same time and to reduce the state complexity.

2. Embedding request has not handled single_queue_is_busy well, line 456 and 466.
3. Check model loaded failed but still proceed for chat/completion.

Next action:
- Replace Serial Queue = Concurrent Queue with workers number = n_parallel
- Wrap embedding request handler within the shared queue

## Issues
- #430

### DIFF
--- a/controllers/llamaCPP.cc
+++ b/controllers/llamaCPP.cc
@@ -1,10 +1,28 @@
 #include "llamaCPP.h"
+
+#include <trantor/utils/SerialTaskQueue.h>
+
 #include "llama.h"
 #include "log.h"
 #include "utils/nitro_utils.h"
 
 using namespace inferences;
 using json = nlohmann::json;
+
+/**
+ * Queue to handle the inference task, this is to ensure that the inference
+ * task is handled in a sequential manner
+ */
+static trantor::SerialTaskQueue queue("worker");
+
+/**
+ * The state of the inference task
+ */
+enum InferenceStatus {
+  PENDING,
+  RUNNING,
+  FINISHED
+};
 
 /**
  * There is a need to save state of current ongoing inference status of a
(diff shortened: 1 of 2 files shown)


controllers/llamaCPP.cc
```diff
@@ -363,31 +353,29 @@ void llamaCPP::inferenceImpl(
...
-        LOG_INFO << "Failing retrying now";
         return nRead;
       }
-      state->is_streaming = false;
       state->instance->single_queue_is_busy = false;
       return 0;
     };
+    // Wait for the single queue to be released
+    // if n_parallel = 1 && single queue is busy
+    while (state->instance->llama.params.n_parallel == 1 &&
```
louis-jan, line 365: Is there any away to get rid of this while loop (timeout?)

tikikun (approved):
LGTM

PR #433, opened 2024-02-14
refactor: simplify state with queued system

louis-jan:
## Description
- Deprecate shared single_queue_busy state & SerialTaskQueue since it would not work with cont_batching = true and n_parallel > 1.
- Deprecate instance's is_stopped state, better using inferenceStatus enum.
- Introduce ConcurrentTaskQueue to handle tasks with the workers number = n_parallel.
- No complicated state handling, put into the queue and it just work (inference & embedding).

Test results
- n_parallel = 1 & cont_batching = false
  ```
  |        T1          |       T2           |       T3          |       T4          |
  |--------------------|--------------------|-------------------|-------------------|
  | Send chat comp.    | Send chat          | A1 completed      | A2 completed      |
  | request A1         | comp. request A2   | A2 started        |                   |
  |                    |                    |                   |                   |
  |                    |                    |                   |                   |
  ```
- n_parallel = 2 & cont_batching = true
  ```
  |      T1      |      T2      |       T3       |       T4       |                T5               |
  |--------------|--------------|----------------|----------------|--------------------------------|
  | Send chat    | Send chat    | A1 & A2 are    | Send chat      | A1 & A2 are                    |         
  | completion   | completion   | running        | completion     | completed                      |                            
  | request A1   | request A2   |                | request A3     | A3 started running             |                 
  ```

### DIFF
--- a/controllers/llamaCPP.cc
+++ b/controllers/llamaCPP.cc
@@ -1,19 +1,11 @@
 #include "llamaCPP.h"
 
-#include <trantor/utils/SerialTaskQueue.h>
-
 #include "llama.h"
 #include "log.h"
 #include "utils/nitro_utils.h"
 
 using namespace inferences;
 using json = nlohmann::json;
-
-/**
- * Queue to handle the inference task, this is to ensure that the inference
- * task is handled in a sequential manner
- */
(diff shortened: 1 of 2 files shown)


tikikun (approved):
LGTM

PR #434, opened 2024-02-15
chore: release request upon completion

louis-jan:
## Describe your changes
- `request_cancel` should be called on done.
- error does not goes through callback


tikikun (approved):
LGTM

PR #439, opened 2024-02-18
Package cuda libs in the archive

A2va:
With this change, users won't need to install the entire CUDA Toolkit.
On Linux, the user will still have to modify LD_LIBRARY_PATH to add the path to nitro.On Windows I'm not sure if this is correct files especially for CUDA v11, so maybe a test release can be good.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -285,6 +285,7 @@ jobs:
         run: |
           mkdir -p nitro
           cp build/nitro nitro/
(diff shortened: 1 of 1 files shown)


hiento09:
Hi @A2va,, we tried this method before, and it did not cover all use cases. The reason is that the CUDA driver comes with a list of compatible NVIDIA drivers. You can refer to it here for Linux: https://docs.nvidia.com/deploy/cuda-compatibility/#use-the-right-compat-package (I have not seen NVIDIA publish this for Windows).
Thank you for taking the time for Nitro!

PR #444, opened 2024-02-22
refactor: modules extends base classes - comply google style guide

louis-jan:
## Descriptions

This pull request refactors whisperCPP & llamaCPP to extend base classes, also comply google style guide

- [x] Extends base class
- [x] Comply google style guide
- [ ] Add model load state & api
- [ ] Clean code

## Issues
- #427

PR #450, opened 2024-02-26
Resolve race condition during initial model load

BachDao:
- The initial client establishes a connection with the server and initiates a model load operation.
- During the loading process, a second client connects and observes that the "model_loaded_external" flag remains false, prompting it to attempt loading the model again, leading to a race condition.


louis-jan:
Great find!

controllers/llamaCPP.cc
```diff
@@ -536,31 +536,40 @@ void llamaCPP::modelStatus(
...
+
+  if(llama.model_loaded_external.load(std::memory_order_acquire)){
+    model_loaded_response(req, callback);
     return;
   }
+  bool modelLoadedSuccess;
+  const auto& jsonBody = req->getJsonObject();
+
+  {
+    std::scoped_lock lck{load_model_mutex};
```
louis-jan, line 548: lock_guard is good enough with this single lock

louis-jan:
I don't know why it's stuck at loading the model on Mac; I'm switching to Windows to test again.

controllers/llamaCPP.cc
BachDao: Hi,for this specific case, both work fine.
In general, if you don't have compatibility requirements, my suggestion is stick with scoped version, better api, better name and one-less-thing to remember.

BachDao:
The last write to the flag should be write_released instead of relaxed, will update when I get back to my laptop.

Anw, if your Windows running x86, it should be fine, otherwise problem in somewhere else.

louis-jan:
> The last write to the flag should be write_released instead of relaxed, will update when I get back to my laptop.
> 
> Anw, if your Windows running x86, it should be fine, otherwise problem in somewhere else.

Great, thank you! Let me know when you need me to test again.

controllers/llamaCPP.cc
```diff
@@ -654,7 +663,6 @@ bool llamaCPP::loadModelImpl(std::shared_ptr<Json::Value> jsonBody) {
 
   queue = new trantor::ConcurrentTaskQueue(llama.params.n_parallel, "llamaCPP");
 
-  llama.model_loaded_external = true;
```
louis-jan, line 657: The hang issue is due to this, @BachDao. There is a run loop in backgroundTask for updating results  Without `update_slots` being executed, `next_result` would wait until then (another run loop and mutex).

controllers/llamaCPP.h
```diff
@@ -2277,5 +2278,8 @@ class llamaCPP : public drogon::HttpController<llamaCPP>, public ChatProvider {
   void warmupModel();
   void backgroundTask();
   void stopBackgroundTask();
+  void model_loaded_response(
```
louis-jan, line 2281: I am refactoring the coding conventions of these controllers. Please refer to issue #444, where it is mentioned that class methods should follow PascalCase.

controllers/llamaCPP.cc
BachDao: Why the background task need to check that flag?
I assume that before you submit any task to background worker, everything should be initialized, or I missing smt ?

controllers/llamaCPP.cc
```diff
@@ -683,3 +691,13 @@ void llamaCPP::stopBackgroundTask() {
     }
   }
 }
+void llamaCPP::model_loaded_response(
+    const HttpRequestPtr ptr,
```
louis-jan, line 695: ptr is not used, remove out

controllers/llamaCPP.cc
BachDao: if you set for C++11, then lock_guard is only choice, totally agree.

louis-jan: Thank you

controllers/llamaCPP.cc
louis-jan: I don't like this check flag either, but the `warmupModel` should not affect the `loadModel` request. That's why the flag is flipped before. There is another clue: when running the model warm-up process, there could also be an unload model request, and in this case, the flag check works (to prevent slot update). However, I'm not a fan of this logic; will try to refactor it somehow soon.

BachDao: Not sure if this is the appropriate place to discuss but I put a few lines below:

- As my understanding, `warmupModel` is called only once for each time model is loaded. If that correct, should both functions be placed on the same thread and run sequentially without synchronization ?
- Regarding `model_loaded_external`, it seems that the flag is used both to denote a successful model load and the continuation of the background task, should it be separated ?
https://github.com/janhq/nitro/blob/2242b7da8ef741e414f70a6be15128a7318186e6/controllers/llamaCPP.cc#L666-L669
- Should the task queue be wrapped behind a simple interface that allows submission, polling, and blocking-wait for results? The current implementation scattered in the controller class makes it a bit hard to follow. Also, having a unified interface provides flexibility for adapting to other use cases if necessary.
-  Will the following block the event loop? Should we consider submitting the continuation and freeing the I/O thread for another request?
https://github.com/janhq/nitro/blob/2242b7da8ef741e414f70a6be15128a7318186e6/controllers/llamaCPP.cc#L419-L421

Putting all these detail aside, it's still not clear to me what the use-cases of Nitro are:

- Is it intended to be used as an embedded server (single-user)? In that case, would a single I/O thread be sufficient? With current implementation of Drogon, for each I/O thread, there is separate event loop that keeps polling the Acceptor's socket, it's too much for single connection.
- Otherwise, if we intend to server multiple client,  the current implementation ties the lifetime of `llama_server_context` to the `controller` object, which has only single-instance for the entire app lifetime.
https://github.com/janhq/nitro/blob/2242b7da8ef741e414f70a6be15128a7318186e6/controllers/llamaCPP.h#L2250-L2251
It mean single inference context will be shared among all client, seem not appropriate. Imagine a scenario where the second client attempts to load a model with specific config but loses in the race, and the configuration comes from the first client.

Regardless, you guys done an excellent job making local inference easy for everyone. Looking forward to your future releases.

louis-jan:
@BachDao request for help, could you please help rebase the latest main?

controllers/llamaCPP.cc
BachDao: - `warmupModel` is a background task...
- Agreed, as above. Should decouple the states...
- Sorry, there are many tasks mentioned here....
**_===> For all the points above, I will try to find sometime and come back with concrete code, it will be easier for discussion._**
-  This is not blocking the event loop; you can see it runs in another thread **_==> agree, the task itself running on background thread, but `next_result(task_id)`, which called on i/o thread can be blocking_wait if task not yet completed._**
- ... "Current implementation of Drogon" Are you referring to the SSE impl? **_==> Drogon, the HttpServer framework that used by Nitro._**

tikikun:
stale

PR #460, opened 2024-03-05
Remove always() condition in run github action jobs and remove e2e testing that hang the CI

hiento09:
Remove always() condition in run github action jobs and remove e2e testing that hang the CI


louis-jan (approved):
LGTM

Issue #463, opened 2024-03-14
Nitro release AVX and AVX512


hiro-v:
Link
https://github.com/janhq/nitro/issues/322

PR #464, opened 2024-03-14
feat: Add Nitro support for different CPU instructions and architecture

hiro-v:
Fix for https://github.com/janhq/nitro/issues/463

- [x] Rename `amd64` to `amd64-avx2` (as default option)
- [x] Add `amd64-avx` (with/ without CUDA)
- [x] Add `amd64-avx512` (with/ without CUDA)
- [x] Refactor `amd64-vulkan` in matrix strategy


hiento09 (approved):
lgtm

MauroDruwel:
Hi,

I think this pull request broke the instalation script... Version 0.3.15 isn't installing on my linux machine and windows machine... Could this be the problem?

Issue #465, opened 2024-03-20
bug: Incorrect content-type in chat completions response


dpkirchner:
**Describe the bug**
The `content-type` header returned on the /v1/chat/completions route is `text/html; charset=utf-8` instead of `application/json`, which causes errors when using clients that use the header to decode responses.

**To Reproduce**
1. Load a model into nitro, any model will do
2. Hit the route with curl: `curl -vv -s http://localhost:3928/v1/chat/completions -H 'content-type: application/json' -d '{"messages":[{"role":"user","content":"Whatever"}]}'`
3. See the incorrect `content-type`

**Expected behavior**
The `content-type` header in the http response should be `application/json`.

**Additional context**

```bash
curl -vv -s http://localhost:3928/v1/chat/completions -H 'content-type: application/json' -d '{"messages":[{"role":"user","content":"Whatever"}]}'
```

output:
```
*   Trying 127.0.0.1:3928...
* Connected to localhost (127.0.0.1) port 3928 (#0)
> POST /v1/chat/completions HTTP/1.1
> Host: localhost:3928
> User-Agent: curl/7.74.0
> Accept: */*
> content-type: application/json
> Content-Length: 51
>
* upload completely sent off: 51 out of 51 bytes
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< content-length: 309
< content-type: text/html; charset=utf-8
< server: drogon/1.9.2
< access-control-allow-origin: *
< date: Wed, 20 Mar 2024 22:58:27 GMT
<
* Connection #0 to host localhost left intact
{"choices":[{"finish_reason":null,"index":0,"message": [snipped]```

louis-jan:
It's addressed. Could you please try again @dpkirchner?

dpkirchner:
lgtm, thanks!

PR #466, opened 2024-03-20
fix: Use json response method for chat completion responses (#465)

dpkirchner:


tikikun:
This will make more sense if added into nitro utils folder @louis-jan can you help to check if we already have application/json somewhere else also?

tikikun:
I will make changes directly in utils

PR #471, opened 2024-03-26
Feature universal bin file macos

hiento09:
Feature universal bin file macos


namchuai (approved):

Van-QA:
linking to [#2068](https://github.com/janhq/jan/issues/2068)

PR #478, opened 2024-04-03
feat: set content type string for json properly

tikikun:


louis-jan:
Also remove `resp->setContentTypeString("application/json");` from `llamaCPP::EmbeddingImpl` @tikikun

PR #486, opened 2024-04-06
Add AVX support for Linux in `install.sh`

MauroDruwel:
This pull request addresses an issue from pull request #464, which caused Nitro software installation failures due to a missing AVX version specification.

I've updated the script to automatically detect the best AVX version using `/proc/cpuinfo` if you are on Linux. I tested this, and it is working on my Linux machine and MacBook. (I also did something similar in pull request #469, but that one was for Windows 😉)

Feel free to contact me if you have any issues! 😄

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,18 @@ check_install_jq_tar() {
     fi
 }
 
+determine_avx_support() {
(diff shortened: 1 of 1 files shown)


hiento09 (approved):
lgtm

PR #488, opened 2024-04-09
fix: responses from /chat/completions endpoint contain a leading space in the content

vansangpfiev:
Issue link: https://github.com/janhq/jan/issues/2548
- Add google test with sample
- Trim the leading space if not in stream mode or it is the first token in stream mode


controllers/llamaCPP.cc
```diff
@@ -330,29 +332,34 @@ void llamaCPP::InferenceImpl(
...
-        LOG_WARN_REQUEST(request_id) "Connection closed or buffer is null. Reset context";
+        LOG_WARN_REQUEST(request_id)
+        "Connection closed or buffer is null. Reset context";
         state->inference_status = FINISHED;
         return 0;
       }
 
       if (state->inference_status == EOS) {
         LOG_INFO_REQUEST(request_id) << "End of result";
+        is_first_token = true;
```
CameronNguyen130820, line 354: Hi @vansangpfiev, can I have a question?
Do we really need to `ltrim` the last token?

vansangpfiev: No, we only reset `is_first_token` state for the last token

vansangpfiev:
Nitro logs:
Call:
```
POST http://127.0.0.1:3928/v1/chat/completions
**headers**
content-type: application/json
**body**
{
    {
      "content": "Hello!",
      "role": "user"
    }
  ],
  "model": "mistral-7b-instruct-v0.1.Q5_0",
  "stream": true,
  "max_tokens": 4096,
  "stop": [
    "</s>"
  ],
  "frequency_penalty": 0,
  "presence_penalty": 0,
  "temperature": 0.7,
  "top_p": 0.95
}
```
Response: 
```
data: {"choices":[{"delta":{"content":"I"},"finish_reason":null,"index":0}],"created":1712803949,"id":"KqHSWp7mbWJpCveUZhve","model":"_","object":"chat.completion.chunk"}
data: {"choices":[{"delta":{"content":"'"},"finish_reason":null,"index":0}],"created":1712803950,"id":"kmChXtpgvhb8AEVIIDQ1","model":"_","object":"chat.completion.chunk"}
data: {"choices":[{"delta":{"content":"m"},"finish_reason":null,"index":0}],"created":1712803951,"id":"bA3EIN56gaTXLLoFnLpv","model":"_","object":"chat.completion.chunk"}
data: {"choices":[{"delta":{"content":" just"},"finish_reason":null,"index":0}],"created":1712803951,"id":"uQ5lSv9u2E4GUCdUejk7","model":"_","object":"chat.completion.chunk"}
data: {"choices":[{"delta":{"content":" a"},"finish_reason":null,"index":0}],"created":1712803952,"id":"WhCvljFp5UFk1qjDgwmZ","model":"_","object":"chat.completion.chunk"}
...
```

controllers/llamaCPP.cc
```diff
@@ -328,29 +330,34 @@ void llamaCPP::InferenceImpl(
   if (is_streamed) {
     LOG_INFO_REQUEST(request_id) << "Streamed, waiting for respone";
     auto state = create_inference_state(this);
+    bool is_first_token = true;
```
tikikun: Should store in state since state is a throwaway object that's automatically managed by shared ptr for one single session

tikikun: @vansangpfiev

tikikun (approved):

controllers/llamaCPP.cc
vansangpfiev: good point, moved it to the state, thanks

Issue #467, opened 2024-03-21
epic: Load multiple models


tikikun:
**Problem**
Right now nitro only can load one model at a time

**Success Criteria**
We need to be able to load multiple models

louis-jan:
It's completed (tested)

PR #495, opened 2024-04-10
feat: load multiple models

vansangpfiev:
Issue link: https://github.com/janhq/nitro/issues/467
- use std::unordered_map to store all llama_server_context (lsc)
- refactor: move background thread to lsc, each lsc has its own background thread
- TODO: update documentation

We use mode_id as a key to find the model, so they have to be unique. That requires some changes in request parameter:
For completions:
Have to set value the model_alias for `inferences/llamaCPP/loadmodel` and `inferences/llamaCPP/unloadmodel` , this has to be the same as model parameter in `inferences/llamaCPP/chat_completion` 
loadmodel/unloadmodel:
```
{
...
 "llama_model_path": "file_to_location"
  "model_alias": "model1"
...
}
```
chat_completion:
```
{
...
  "model": "model1"  
...
}
```
For embeddings:
The same for loadmodel/unloadmodel, for embedding request, we need to add model parameter
loadmodel/unloadmodel
```
{
...
    "llama_model_path": "e:/workspace/model/nomic-embed-text-v1.5.f16.gguf",
    "model_alias": "model1",
    "model_type": "embedding"
...
}
```
embedding
```
{
...  
  "input": "how are you",
  "model": "model1"
...
}
```

### DIFF
--- a/controllers/llamaCPP.cc
+++ b/controllers/llamaCPP.cc
@@ -28,11 +28,11 @@ enum InferenceStatus { PENDING, RUNNING, EOS, FINISHED };
 struct inferenceState {
   int task_id;
   InferenceStatus inference_status = PENDING;
-  llamaCPP* instance;
+  llama_server_context& llama;
   // Check if we receive the first token, set it to false after receiving
   bool is_first_token = true;
 
-  inferenceState(llamaCPP* inst) : instance(inst) {}
+  inferenceState(llama_server_context& l) : llama(l) {}
 };
 
 /**
@@ -40,8 +40,9 @@ struct inferenceState {
  * inferenceState will be persisting even tho the lambda in streaming might go
  * out of scope and the handler already moved on
  */
-std::shared_ptr<inferenceState> create_inference_state(llamaCPP* instance) {
-  return std::make_shared<inferenceState>(instance);
+std::shared_ptr<inferenceState> create_inference_state(
+    llama_server_context& l) {
+  return std::make_shared<inferenceState>(l);
 }
 
 /**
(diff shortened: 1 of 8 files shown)


controllers/llamaCPP.cc
```diff
@@ -681,62 +705,47 @@ bool llamaCPP::LoadModelImpl(std::shared_ptr<Json::Value> jsonBody) {
...
-  if (queue != nullptr) {
-    delete queue;
+  server_ctx_map[model_id].initialize();
+
+  if (inference_task_queue == nullptr ||
+      task_queue_thread_num < params.n_parallel) {
+    task_queue_thread_num = std::max(task_queue_thread_num, params.n_parallel);
+    LOG_INFO << "Start inference task queue, num threads: "
+             << task_queue_thread_num;
+    inference_task_queue = std::make_unique<trantor::ConcurrentTaskQueue>(
```
louis-jan, line 731: The inference task queue is constructed with the model's parameters. It is shared among the model's operations, which is incorrect, right? Let's say it would be renewed every time a new model request with a larger n_parallel number comes in:
1. How will pending tasks be processed since it is renewed?
2. Will later requests change thread_num, causing side effects for previously loaded models?"

utils/nitro_utils.h
```diff
@@ -282,4 +283,31 @@ inline void ltrim(std::string& s) {
...
+  } else if(!jsonBody["model_alias"].isNull()) {
+    return jsonBody["model_alias"].asString();
+  }
+  
+  // We check llama_model_path for loadmodel request
+  if (auto input = jsonBody["llama_model_path"]; !input.isNull()) {
+    auto s = input.asString();
+    std::replace(s.begin(), s.end(), '\\', '/');
+    auto const pos = s.find_last_of('/');
+    // We only if file name has gguf extension or nothing
```
louis-jan, line 299: Missing some words

controllers/llamaCPP.cc
vansangpfiev: One solution (though it is not optimal) is that we can use a task queue for each model. Will change the code to make it works correctly first.

louis-jan (approved):
LGTM

vansangpfiev:
Moved PR to https://github.com/janhq/cortex.llamacpp/pull/14

PR #498, opened 2024-04-11
fix: make embedding work again

vansangpfiev:
Changes in llama.cpp require our interface changes.
To load the embedding model, need to set embedding flag to true, for example:
```
{
    "llama_model_path": "e:/workspace/model/nomic-embed-text-v1.5.f16.gguf",
    "ctx_len": 5,
    "ngl": 0,
    "n_parallel": 1,
    "embedding": true
} 
```

### DIFF
--- a/context/llama_server_context.h
+++ b/context/llama_server_context.h
@@ -1,15 +1,15 @@
+#include <mutex>
+#include <set>
 #include <string>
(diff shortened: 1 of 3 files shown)


tikikun (approved):
LGTM

PR #504, opened 2024-04-12
fix: embedding fixup for nomic-embed-text model

vansangpfiev:
For model like nomic-embed-text-v1.5.f16.gguf, we skip warm up.
Also change the API to send_embedding for non-logits model.

### DIFF
--- a/controllers/llamaCPP.cc
+++ b/controllers/llamaCPP.cc
@@ -156,7 +156,7 @@ llamaCPP::~llamaCPP() {
   StopBackgroundTask();
 }
 
-void llamaCPP::WarmupModel(bool is_embedded_model) {
+void llamaCPP::WarmupModel() {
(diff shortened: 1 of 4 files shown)


controllers/llamaCPP.cc
```diff
@@ -649,6 +654,8 @@ bool llamaCPP::LoadModelImpl(std::shared_ptr<Json::Value> jsonBody) {
 
     params.n_gpu_layers = jsonBody->get("ngl", 100).asInt();
     params.n_ctx = jsonBody->get("ctx_len", 2048).asInt();
+    is_embedded_model =
```
vansangpfiev, line 657: use embedding flag to check if we should warm up model with embedding or not

tikikun (approved):
LGTM

PR #510, opened 2024-04-16
fix: reject completion request for embedding model

vansangpfiev:
Embedding model does not support completion, so should reject completion request.


tikikun (approved):
LGTM

PR #513, opened 2024-04-16
fix: app hangs if we unload model during stream completion 

vansangpfiev:
Currently e2e scripts can be hang because we unload model when still receiving the stream. Change to non-stream to temporary fix issue.


tikikun:
If it hangs when unloading model it also signifies another bug

tikikun (approved):
LGTM

PR #529, opened 2024-04-23
fix: ci for windows cuda builds

vansangpfiev:
Has some errors when we move the windows cuda build to new machine.


hiento09 (approved):
Lgtm

PR #532, opened 2024-04-24
fix: Update CI for build

hiro-v:
- [x] Update `nitro-set-version` with cortex
- [x] Remove metal file in mac-arm64

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -98,7 +98,7 @@ jobs:
               local max_retries=3
               local tag
               while [ $retries -lt $max_retries ]; do
-                tag=$(curl -s https://api.github.com/repos/janhq/nitro/releases/latest | jq -r .tag_name)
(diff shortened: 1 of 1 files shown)


hiento09:
@hiro-v , can you help to change `macos-latest` to `macos-13` for macintel build, since github just change label `macos-latest` from m1 to intel yesterday, which cause this step error 
image

.github/workflows/build.yml
hiento09, line 1: I have checked the job at  https://github.com/ggerganov/llama.cpp/actions/runs/8635935781/job/23674807267#step:5:2313. The tests failed because of this flag: `-DLLAMA_METAL_EMBED_LIBRARY=ON`. We don't have it enabled on Mac x64, and I don't think we need to add this flag for Mac x64 since it's not enabled by default."

hiento09 (approved):

PR #538, opened 2024-05-03
feat: llama engine

vansangpfiev:
- Remove llama.cpp submodule
- Remove whisper.cpp submodule
- Add engine.cmake to pull cortex.llamacpp stable release
- Update build.yml to pack cortex.llamacpp library as well 
TODO: rename nitro to cortex.cpp (?) in another PR


louis-jan (approved):
LGTM

PR #542, opened 2024-05-06
feat: add command support

namchuai:


cortex-js/.env.development
```diff
@@ -0,0 +1,7 @@
+# Directories
+## The directories that contain the models
+CORTEX_MODELS_DIR=/Users/jamesnguyen/models
```
namchuai, line 3: Remember to update with your corresponding path

PR #558, opened 2024-05-14
feat: change unload model and model status to POST

vansangpfiev:
cortex.llamacpp is now supporting multiple models, we need to change method of `unloadmodel` and `modelstatus` to POST.
Sample requests for multiple models feature:
load model:
```
curl http://localhost:3928/inferences/server/loadmodel \
  -H 'Content-Type: application/json' \
  -d '{
    "llama_model_path": "/model/llama-2-7b-model.gguf",
    "model_alias": "llama-2-7b-model",
    "model_type": "llm" // use 'embedding` for embedding model
  }'
``` 
unload model
```
curl http://localhost:3928/inferences/server/unloadmodel \
  -H 'Content-Type: application/json' \
  -d '{
    "model": "llama-2-7b-model"
  }'
```


louis-jan (approved):
LGTM

PR #559, opened 2024-05-14
feat: python runtime engine

vansangpfiev:


cortex-cpp/main.cc
```diff
@@ -18,6 +20,27 @@
...
+        return 1;
+      }
+
+      auto func = dl->get_function<CortexPythonEngineI*()>("get_engine");
+      auto e = func();
+      e->ExecutePythonFile(argv[0], argv[2], py_home_path);
+      return 0;
+    }
+  }
+
```
louis-jan, line 43: As I remember it's a test script, we can remove it out in prod

vansangpfiev: No, it's not. In reality, we create another process to run the fine-tune job with parameters include `python-file` (which is fine-tune script)

louis-jan: I didn't see the location; this is actually the main entry of the engine, not the program.

louis-jan (approved):
LGTM

PR #570, opened 2024-05-15
feat: cpu instruction set detection

vansangpfiev:
Check the CPU instruction set supported before loading the engine.
Only LOGWRN if it is not compatible for now.


louis-jan (approved):
LGTM

PR #577, opened 2024-05-16
chore: separating commands

namchuai:


louis-jan:
Can't stop cortex-cpp process, it hang and stop working after the first run

louis-jan:
Let's merge then fix it in another PR since this is a structure update PR

PR #592, opened 2024-05-21
feat: get running models

vansangpfiev:
issue: https://github.com/janhq/cortex/issues/589
endpoint: v1/models
response example:
```
{
    "data": [
        {
            "id": "nomic-embed-text-v1.5.f16",
            "object": "model"
        },
        {
            "id": "mistral-7b-instruct-v0.1.Q4_0",
            "object": "model"
        }
    ],
    "object": "list"
}
```

// TODO: bump cortex.llamacpp version to support this


louis-jan (approved):
LGTM

PR #633, opened 2024-05-30
fix: allow users to select a model to start

louis-jan:
## Describe Your Changes

- This PR is to support selecting a model prompt when running `cortex models start`
- 
Screenshot 2024-05-30 at 13 18 51

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/cortex-js/src/infrastructure/commanders/models/model-start.command.ts
+++ b/cortex-js/src/infrastructure/commanders/models/model-start.command.ts
@@ -1,4 +1,9 @@
-import { CommandRunner, SubCommand, Option } from 'nest-commander';
+import {
+  CommandRunner,
+  SubCommand,
+  Option,
+  InquirerService,
+} from 'nest-commander';
 import { exit } from 'node:process';
 import { ModelsCliUsecases } from '../usecases/models.cli.usecases';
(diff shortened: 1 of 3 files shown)


cortex-js/src/infrastructure/commanders/models/model-start.command.ts
```diff
@@ -9,16 +14,22 @@ type ModelStartOptions = {
...
 
   async run(input: string[], options: ModelStartOptions): Promise<void> {
-    if (input.length === 0) {
-      console.error('Model ID is required');
-      exit(1);
+    let modelId = input[0];
+    if (!modelId) {
+      try {
+        modelId = await this.modelInquiry();
+      } catch {
```
namchuai, line 29: maybe just log the error?

namchuai (approved):

cortex-js/src/infrastructure/commanders/models/model-start.command.ts
louis-jan: No, can't log error here. 

This is a fallback for the case modelId is not inputted. Not a happy case to log weird error out.

PR #642, opened 2024-06-02
feat: cortex post-install script - init silently on install

louis-jan:
## Describe Your Changes
- This PR adds a post-install script that downloads the corresponding Cortex dependencies based on the user's hardware specs during installation. So, after installing Cortex from the npm repo, there is no need to run `cortex init` to get started.

- Removed cortex init step from README

Support: 
- Nvidia GPUs (ROCm is in progress)
- CPU with AVX-512, AVX2, AVX
- Apple Silicon

image

## Self Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,12 +68,7 @@ To install Cortex CLI, follow the steps below:
 npm i -g @janhq/cortex
 ```
 
-2. Initialize a compatible engine:
-``` bash
-cortex init
-```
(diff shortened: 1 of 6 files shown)


namchuai (approved):

namchuai:
LGTM. We might need to take care of the case user turn on/off AVX in the future.

PR #709, opened 2024-06-17
chore: update models to extends from openai

namchuai:
## Describe Your Changes

- Update object model to extends from OpenAI

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


cortex-js/src/domain/models/model.interface.ts
```diff
@@ -1,8 +1,29 @@
...
- * @stored
- */
-export interface Model {
+import { Model as OpenAiModel } from 'openai/resources/models';
+
+export const LocalEngines = [
+  'cortex.llamacpp',
+  'cortex.onnx',
+  'cortex.tensorrt',
+] as const;
```
louis-jan, line 7: This is not correct

cortex-js/src/infrastructure/commanders/usecases/chat.cli.usecases.ts
```diff
@@ -76,22 +72,22 @@ export class ChatCliUsecases {
 
       messages.push({
         content: userInput,
-        role: ChatCompletionRole.User,
+        role: 'user',
```
louis-jan, line 75: why changed from enum to text?

louis-jan: because the enum is not well annotated?

cortex-js/src/domain/models/model.interface.ts
namchuai: removed

cortex-js/src/infrastructure/commanders/usecases/chat.cli.usecases.ts
namchuai: hmm, just following openai's type.

louis-jan:
Rebase needed @namchuai

louis-jan (approved):

PR #818, opened 2024-07-03
chore: error output logs for cortex run

### LINKED ISSUE #776: feat: Better stdout when pulling models
**Current:** 
Model __ not found. Try pulling model ...

**Problem:** 
1. Sounds like the model isn't available (for download). 
1. Try pulling model sounds like you are telling the user to manually do it.

**Revision:** 
__ not found on filesystem. 
Downloading from remote: https://(insert the url, if possible).

### LINKED ISSUE #801: feat: stdout "loading model..." during `run` and `start`
When running any model, it takes a while to load model. It looks like the command didn't work, leading users to ctrl-c or be confused. We should stdout a transition message like “loading model...”

louis-jan:
chore: error handling - cortex run error outputs and logs file

## Describe Your Changes

Added error outputs and log file

NOTES: Removed `Checking model files for existing...`

`cortex run`: Loading status

image

`cortex run`: Error outputs from cortex-js

image

`cortex run`: Error outputs from cortex-cpp

image

`cortex pull`: with remote URL 

image

image

## Fixes Issues

- Closes #801
- Closes #776

## Self Checklist

### DIFF
--- a/cortex-js/src/utils/logs.ts
+++ b/cortex-js/src/utils/logs.ts
@@ -1,0 +1,32 @@
+import { createReadStream } from 'fs';
+import { join } from 'path';
+import { createInterface } from 'readline';
+
+/**
+ * Print the last N lines of a file that contain the word 'ERROR'
+ * @param filename
+ * @param numLines
+ */
+export async function printLastErrorLines(
+  dataFolderPath: string,
+  numLines: number = 5,
+): Promise<void> {
+  const errorLines: string[] = [];
+
+  const fileStream = createReadStream(join(dataFolderPath, 'cortex.log'));
+  const rl = createInterface({
+    input: fileStream,
+    crlfDelay: Infinity,
+  });
+
+  for await (const line of rl) {
+    errorLines.push(line);
+    if (errorLines.length > numLines) {
+      errorLines.shift();
+    }
+  }
+
+  console.log(`Last errors:`);
+  errorLines.forEach((line) => console.log(line));
+  console.log('...');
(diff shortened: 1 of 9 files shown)


louis-jan:
Removed `Checking model files for existing...`

cortex-js/src/utils/logs.ts
```diff
@@ -0,0 +1,32 @@
...
+import { createInterface } from 'readline';
+
+/**
+ * Print the last N lines of a file that contain the word 'ERROR'
+ * @param filename
+ * @param numLines
+ */
+export async function printLastErrorLines(
+  dataFolderPath: string,
+  numLines: number = 5,
```
gabrielle-ong, line 12: stupid qn from me - any reason why only 5 lines (will they see a truncated error message?)

gabrielle-ong: non blocking curious qn, otherwise LGTM!

gabrielle-ong (approved):

cortex-js/src/utils/logs.ts
louis-jan: > stupid qn from me - any reason why only 5 lines (will they see a truncated error message?)

Yeah, that's a rough estimate for common cases. Since we output everything into the log file across sessions, we should output only several lines (not just errors), similar to other applications. Otherwise, there would be hundreds of lines of cortex-cpp printed out. The applications are not running in attached mode; in attached mode, users would see the full session logs.

PR #891, opened 2024-07-18
feat: support running multiple engines at the same time

vansangpfiev:
## Describe Your Changes

- Currently we only allow to run one engine at a time. Because cortex.trt-llm and cortex.onnx don't support embeddings yet, therefore we need to run cortex.llamacpp for embeddings along side with cortex.trt-llm/cortex.onnx for chat completions.
- This PR also includes a commit to refactor the codebase.


louis-jan (approved):
LGTM

PR #982, opened 2024-08-06
fix: add dll search path

### LINKED ISSUE #1047: feat: Set dll libraries search path for Cortex on Windows
Our current approach is copy all .dll files into execution folder. We support multiple engines and if we keep dll files in `engines/cortex.xxx` folder, it seems cleaner. Should consider to find a solution to set dll search path.

vansangpfiev:
## Describe Your Changes

- On Windows, we can add dll directory to search libraries on runtime https://learn.microsoft.com/en-us/windows/win32/api/libloaderapi/nf-libloaderapi-adddlldirectory.

## Fixes Issues

- Closes https://github.com/janhq/cortex/issues/1047


namchuai (approved):
✅ Great work!

engine/commands/engine_init_cmd.cc
```diff
@@ -224,15 +197,24 @@ bool EngineInitCmd::Exec() const {
...
-
+                  // TODO(any) This is a temporary fix. The issue will be fixed when we has CIs
+                  // to pack CUDA dependecies into engine release
+                  auto get_engine_path = [](std::string_view e) {
+                    if (e == "cortex.llamacpp") {
+                      return cortex_utils::kLlamaLibPath;
+                    } else if (e == "cortex.tensorrt-llm") {
+                      return cortex_utils::kTensorrtLlmPath;
+                    }
+                    return cortex_utils::kLlamaLibPath;
```
dan-menlo: Lines 208 and 212 - is it extraneous to have 208?

engine/controllers/server.cc
```diff
@@ -290,10 +288,46 @@ void server::LoadModel(const HttpRequestPtr& req,
...
-      std::cout << abs_path << std::endl;
+#if defined(_WIN32)
+      // TODO(?) If we only allow to load an engine at a time, the logic is simpler.
+      // We would like to support running multiple engines at the same time. Therefore,
+      // the adding/removing dll directory logic is quite complicated:
+      // 1. If llamacpp is loaded and new requested engine is tensorrt-llm:
+      // Unload the llamacpp dll directory then load the tensorrt-llm
+      // 2. If tensorrt-llm is loaded and new requested engine is llamacpp:
+      // Do nothing, llamacpp can re-use tensorrt-llm dependencies (need to be tested careful)
+      // 3. Add dll directory if met other conditions
```
dan-menlo, line 303: Can I understand this in more detail: 
- We are always loading the dll directory of the latest used engine
- Is this due to our decision to keep per-engine dependencies?
- Are we able to load dlls on an as-needed basis, without removing the "older" engine dlls? 

I imagine this will come up often in demos - we will demo:
```
cortex run llama3.1
cortex run llama3.1:tensorrt-llm
cortex run llama3.1:onnx
```

dan-menlo:
@vansangpfiev I do not have sufficient knowledge to be able to assess this PR, and have just approved it first to ensure you guys are not blocked. 

I've left a couple of basic clarifying questions - if you don't mind

dan-menlo (approved):

engine/commands/engine_init_cmd.cc
vansangpfiev: > Lines 208 and 212 - is it extraneous to have 208?

We fallback to default, will change the function to make it clearer.

engine/controllers/server.cc
vansangpfiev: We only load the engine's dll directory if engine is not loaded. In case of tensorrt-llm and llamacpp engine, it is a little bit tricky because the cuda dependencies may conflict so we need to remove before add the dll directory.

PR #996, opened 2024-08-08
fix: remove uploads_folder_path

vansangpfiev:
## Describe Your Changes

Upload folder is redundant because we don't upload any files


louis-jan (approved):
LGTM

PR #1029, opened 2024-08-26
feat: pack cortex.llamacpp to cortex-cpp build

vansangpfiev:
## Describe Your Changes

- Pull cortex.llamacpp latest via cmake


hiento09 (approved):
lgtm

PR #1034, opened 2024-08-27
feat: add engine install cli

namchuai:
## Describe Your Changes

- Add engine install with hardware detection.

### Engine init `cortex.llamacpp` on windows. For this machine, AMD Ryzen 5950x, it choose AVX2, cuda version 12.0
image

### Engine init `cortex.onnx`
image

### Engine init `cortex.tensorrt-llm`
image

## Self Checklist

### DIFF
--- a/engine/controllers/engines.cc
+++ b/engine/controllers/engines.cc
@@ -1,0 +1,111 @@
+#include "engines.h"
+#include "utils/archive_utils.h"
+#include "utils/file_manager_utils.h"
+#include "utils/system_info_utils.h"
+
+void Engines::InitEngine(const HttpRequestPtr& req,
+                         std::function<void(const HttpResponsePtr&)>&& callback,
+                         const std::string& engine) const {
+  LOG_DEBUG << "InitEngine, Engine: " << engine;
+  if (engine.empty()) {
+    Json::Value res;
+    res["message"] = "Engine name is required";
+    auto resp = cortex_utils::CreateCortexHttpJsonResponse(res);
+    resp->setStatusCode(k409Conflict);
+    callback(resp);
+    LOG_WARN << "No engine field in path param";
+    return;
+  }
+
+  auto system_info = system_info_utils::GetSystemInfo();
+  if (system_info.arch == system_info_utils::kUnsupported ||
+      system_info.os == system_info_utils::kUnsupported) {
+    Json::Value res;
+    res["message"] = "Unsupported OS or architecture";
+    auto resp = cortex_utils::CreateCortexHttpJsonResponse(res);
+    resp->setStatusCode(k409Conflict);
+    callback(resp);
+    LOG_ERROR << "Unsupported OS or architecture: " << system_info.os << ", "
+              << system_info.arch;
+    return;
+  }
+
+  auto version{"latest"};
+  constexpr auto gitHubHost = "https://api.github.com";
+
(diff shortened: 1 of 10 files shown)


engine/controllers/engines.cc
```diff
@@ -0,0 +1,123 @@
...
+    callback(resp);
+    LOG_ERROR << "Unsupported OS or architecture: " << system_info.os << ", "
+              << system_info.arch;
+    return;
+  }
+
+  using namespace nlohmann;
+
+  // TODO: add more engines
+  if (engine != "cortex.llamacpp") {
```
vansangpfiev: We only support `cortex.llamacpp` for now?

engine/utils/command_executor.h
```diff
@@ -0,0 +1,49 @@
...
+
+    while (fgets(buffer.data(), buffer.size(), m_pipe.get()) != nullptr) {
+      result += buffer.data();
+    }
+
+    return result;
+  }
+
+ private:
+  std::unique_ptr<FILE, decltype(&PCLOSE)> m_pipe{nullptr, PCLOSE};
```
vansangpfiev, line 48: nice!

engine/utils/engine_matcher_utils.h
```diff
@@ -0,0 +1,195 @@
...
+    "cortex.llamacpp-0.1.25-25.08.24-windows-amd64-vulkan.tar.gz",
+};
+const std::vector<std::string> cortex_onnx_variants{
+    "cortex.onnx-0.1.7-windows-amd64.tar.gz"};
+
+const std::vector<std::string> cortex_tensorrt_variants{
+    "cortex.tensorrt-llm-0.0.9-linux-cuda-12-4.tar.gz",
+    "cortex.tensorrt-llm-0.0.9-windows-cuda-12-4.tar.gz"};
+
+inline std::string GetSuitableAvxVariant();
```
vansangpfiev: Can remove this because we've already had the implementation at the same header file

engine/utils/engine_matcher_utils.h
```diff
@@ -0,0 +1,195 @@
...
+                                 const std::string& cpu_arch);
+inline std::string validate_tensorrtllm(
+    const std::vector<std::string>& variants, const std::string& os,
+    const std::string& cuda_version);
+inline std::string validate(const std::vector<std::string>& variants,
+                            const std::string& os, const std::string& cpu_arch,
+                            const std::string& suitable_avx,
+                            const std::string& cuda_version);
+
+inline std::string validate_tensorrtllm(
```
vansangpfiev: naming convention for function, should be `ValidateTensorrtllm`

vansangpfiev:
I think we should change from `init` to `install`
The command will be something like:
```
cortex engines tensorrt-llm install
```

engine/controllers/engines.cc
namchuai: ah, wrong code. We should support all three: `llamacpp`, `onnx`, and `tensorrt-llm`. Let me remove that check

namchuai:
> I think we should change from `init` to `install` The command will be something like:
> 
> ```
> cortex engines tensorrt-llm install
> ```

Hmm, this will need to change a little bit because it goes against the design of cli application. Might have to add multiple subcommand though.

vansangpfiev (approved):

PR #1043, opened 2024-08-28
Feat background process server

### LINKED ISSUE #1070: Start process in background
Before starting to inference, we need to start a server, this server will be started and ran in background

nguyenhoangthuan99:


engine/main.cc
```diff
@@ -20,8 +22,104 @@
 #error "Unsupported platform!"
 #endif
 
+
+void run_server(){
```
vansangpfiev: Please follow naming convention

engine/main.cc
```diff
@@ -44,58 +142,15 @@ int main(int argc, char* argv[]) {
...
-  LOG_INFO << "Server started, listening at: " << host << ":" << port;
-  LOG_INFO << "Please load your model";
-  drogon::app().addListener(host, port);
-  drogon::app().setThreadNum(drogon_thread_num);
-  LOG_INFO << "Number of thread is:" << drogon::app().getThreadNum();
-
-  drogon::app().run();
-
-  return 0;
+  fork_process();
```
vansangpfiev: should have `return 0`?

vansangpfiev (approved):

PR #1051, opened 2024-08-30
fix: use model_id.yaml for model info

vansangpfiev:
https://github.com/janhq/cortex/issues/1052
Some commands are not working because we use model/model.yml file for model info. 
Use model_id.yaml instead to fix the issue.


namchuai (approved):

ux-han:
@vansangpfiev can you give some updates on this

Issue #1087, opened 2024-09-04
epic: `cortex update` command


vansangpfiev:
**Problem**
This ticket is used for command implementation
cortex should be able to update itself
Proposed command: eg for `v0.5.0-120`
```
cortex update -v <version>
```
```
cortex update
```
If version is not specified then pull the latest version

**Additional context**
UX for `cortex update` is discussed in: https://github.com/janhq/cortex/issues/1041

## Features / Tasklist
- [x] Updater for stable release
- [x] #1169
- [x] #1170

## Bugs
- [x] #1350
- [x] #1257 
- [x] #1278
- [x] #1307
- [x] #1311

freelerobot:
See #1311

gabrielle-ong:
QA: `sudo cortex update` and `sudo cortex update -v 0.5.0-124` works
✅ mac
✅ windows
✅ ubuntu

PR #1095, opened 2024-09-04
feat: cortex update command

vansangpfiev:
https://github.com/janhq/cortex.cpp/issues/1087 
Only support to download latest version for now
Also added `cortex -v` command
https://github.com/janhq/cortex.cpp/issues/1066
```
.\cortex.exe -v
v0.5.0
```


namchuai:
Nice one! I have one question. What happen in windows, if the server is running, then our executable will be in used, right? ? will `std::rename` fail and failed the update, I think?

engine/commands/cortex_upd_cmd.h
```diff
@@ -0,0 +1,15 @@
...
+#include <optional>
+
+namespace commands {
+
+class CortexUpdCmd{
+ public:
+  CortexUpdCmd();
+  void Exec(std::string version);
+
+ private:
```
namchuai: small nit: maybe remove `private` if we don't have any field

engine/commands/cortex_upd_cmd.cc
```diff
@@ -0,0 +1,167 @@
...
+  } else {
+    auto err = res.error();
+    CTL_ERR("HTTP error: " << httplib::to_string(err));
+    return;
+  }
+#if defined(_WIN32)
+  std::string temp = ".\\cortex_tmp.exe";
+  remove(temp.c_str());  // ignore return code
+
+  std::string src =
```
namchuai: small nit, apply `std::filesystem::path` for concatenation path will simplify the code.

engine/commands/cortex_upd_cmd.cc
```diff
@@ -0,0 +1,167 @@
...
+      ".\\cortex\\" + kCortexBinary + "\\" + kCortexBinary + ".exe";
+  std::string dst = ".\\" + kCortexBinary + ".exe";
```
namchuai: I think using `file_manager_utils::GetExecutableFolderContainerPath` might be better for storing files relatively to our binary (executable)

namchuai (approved):
✅ Looks good. Just some small nits. Not required though.

nguyenhoangthuan99 (approved):

vansangpfiev:
> Nice one! I have one question. What happen in windows, if the server is running, then our executable will be in used, right? ? will `std::rename` fail and failed the update, I think?

We cannot replace but can remove file on windows if it is in used.
Also added a TODO to stop server if it is running before updating.

dan-menlo:
I am removing this from Project Board:

- Project Board is for issues, issue is alr tracked in #1087 
- Issues should be closed via PR, via close keyword

PR #1120, opened 2024-09-06
Feat uds for log file

### LINKED ISSUE #1056: bug: Multiprocess cannot write log to the same file
**Problem**
In windows, once a file is taken by one process, other process can not modify it. For example, when the server is writing to log file, it hold `cortex.log`, and if we run CLI command, it cannot write message to `cortex.log`.

**Solution**
In window, we are planning to use Unix Domain Socket to start a thread and write log to file. Below are detail implementation:
- When start server, will create a thread to run UDS to listen messages and write to logfile.
- From server side, override log method of Trantor to send log message to UDS server
- From CLI side, when execute command, check if UDS server is live, if not write directly log to log file.
- With command has `--verbose` flag, log everything to terminal.

nguyenhoangthuan99:


dan-menlo:
@nguyenhoangthuan99 I am not merging this yet, as I am not sure if this approach is correct: 

- It seems that we are doing this to bypass a sensible Windows protection (e.g. single process lock on file)
- This seems fragile; 

I have commented in the original Issue: 

https://github.com/janhq/cortex.cpp/issues/1153

- Should we consider separate logfiles for CLI and server, as 2 separate processes?
- This could be a simpler change

Additionally, when you re-submit the PR, please make sure you link the issue https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue

nguyenhoangthuan99:
I think we should close this PR, because we no longer need UDS. I'll update the new logger in another PR

dan-menlo:
@nguyenhoangthuan99 I typically have submitters close their own PR - can you do so when ready?

nguyenhoangthuan99:
I'll close this PR.

PR #1132, opened 2024-09-06
fix: correct path to work with data folder structure

vansangpfiev:
https://github.com/janhq/cortex.cpp/issues/1121
Only solved issue for llamacpp on MacOS and Linux/Windows CPU build
Need https://github.com/janhq/cortex.cpp/issues/1047 to make others work

### DIFF
--- a/engine/main.cc
+++ b/engine/main.cc
@@ -4,7 +4,6 @@
 #include "controllers/command_line_parser.h"
 #include "cortex-common/cortexpythoni.h"
 #include "utils/archive_utils.h"
-// #include "utils/config_yaml_utils.h"
 #include "utils/cortex_utils.h"
 #include "utils/dylib.h"
 #include "utils/file_manager_utils.h"
--- a/engine/main.cc
+++ b/engine/main.cc
@@ -4,7 +4,7 @@
 #include "controllers/command_line_parser.h"
 #include "cortex-common/cortexpythoni.h"
 #include "utils/archive_utils.h"
-#include "utils/config_yaml_utils.h"
+// #include "utils/config_yaml_utils.h"
 #include "utils/cortex_utils.h"
 #include "utils/dylib.h"
 #include "utils/file_manager_utils.h"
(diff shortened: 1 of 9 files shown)


engine/main.cc
```diff
@@ -4,7 +4,7 @@
 #include "controllers/command_line_parser.h"
 #include "cortex-common/cortexpythoni.h"
 #include "utils/archive_utils.h"
-#include "utils/config_yaml_utils.h"
+// #include "utils/config_yaml_utils.h"
```
namchuai: maybe remove it?

engine/utils/config_yaml_utils.h
```diff
@@ -13,7 +13,7 @@ struct CortexConfig {
   std::string port;
 };
 
-const std::string kCortexFolderName = "cortexcpp";
+const std::string kCortexFolderName = ".cortexcpp";
```
namchuai: Hmm, not sure if we want to hide this folder from user.

engine/utils/file_manager_utils.h
```diff
@@ -84,23 +85,54 @@ inline std::filesystem::path GetConfigurationPath() {
...
+inline std::filesystem::path GetCortexDataPath() {
   // TODO: We will need to support user to move the data folder to other place.
   // TODO: get the variant of cortex. As discussed, we will have: prod, beta, nightly
-  // currently we will store cortex data at ~/.cortex
+  // currently we will store cortex data at ~/.cortexcpp
+  auto config = GetCortexConfig();
+  std::filesystem::path data_folder_path;
+  if (!config.dataFolderPath.empty()) {
+    data_folder_path =
+        std::filesystem::path(config.dataFolderPath) / ".cortexcpp";
```
namchuai: since default `dataFolderPath` is having `kCortexFolderName` already, do we need to add another `.cortexcpp` under it?

namchuai (approved):

nguyenhoangthuan99 (approved):

PR #1172, opened 2024-09-09
chore: update `.cortexrc` fields

namchuai:
## Describe Your Changes

- Update the fields from
`host` -> `apiServerHost`
`port` -> `apiServerPort`
Previously I mistakenly put it as `host` and `port`

## Screenshot
Screenshot 2024-09-09 at 10 18 42

## Fixes Issues
- https://github.com/janhq/cortex.cpp/issues/1031

## Self Checklist

### DIFF
--- a/engine/main.cc
+++ b/engine/main.cc
@@ -26,7 +26,7 @@
 
 void RunServer() {
   auto config = file_manager_utils::GetCortexConfig();
-  LOG_INFO << "Host: " << config.host << " Port: " << config.port << "\n";
+  LOG_INFO << "Host: " << config.apiServerHost << " Port: " << config.apiServerPort << "\n";
 
   // Create logs/ folder and setup log to file
   std::filesystem::create_directory(cortex_utils::logs_folder);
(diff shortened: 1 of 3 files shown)


vansangpfiev (approved):

dan-menlo:
@namchuai @vansangpfiev Given the amount of port number collisions at 4 digits, should we consider a standard port for Cortex that is higher rank? 

- 267839 = Cortex in T9

Outside of that: how did we choose 3928?

dan-menlo (approved):
Lgtm

dan-menlo:
@namchuai Out of curiosity, what was the issue with the initial `port` and `host` varnames? 

- I'm not sure if it's an improvement to make it more verbose
- Given that cortex.cpp only has a single port/host now (vs. cortex platform in the original issue)

namchuai:
> @namchuai Out of curiosity, what was the issue with the initial `port` and `host` varnames?
> 
> * I'm not sure if it's an improvement to make it more verbose
> * Given that cortex.cpp only has a single port/host now (vs. cortex platform in the original issue)

One of the reason is I want to make it compatible with what user currently have. Some user already installed cortex (cortexjs) and they already have `.cortexrc`

PR #1173, opened 2024-09-09
feat-logger-update

### LINKED ISSUE #1056: bug: Multiprocess cannot write log to the same file

### LINKED ISSUE #1153: epic: Cortex.cpp logs
## Goal

- Strong logging infra for Server and CLI
- Users can find and submit logfiles easily in case of bugs
- Logs should ideally not have PII for privacy (e.g. if published on Github)

## Open Questions

- [ ] Should we have separate logs for Server and CLI? (remove needs for UDS)
- [ ] What is our log pruning strategy? 

## Tasklist

- [x] #1054 
- [x] https://github.com/janhq/cortex.cpp/issues/1038
- [x] https://github.com/janhq/cortex.cpp/issues/1007
- [x] https://github.com/janhq/cortex.cpp/issues/1062

nguyenhoangthuan99:
Fixes #1153


nguyenhoangthuan99:
- [x] separate log file for `cortex.log` and `cortex-cli.log`
- [x] pruning log file only to save last 100,000 lines. The log file will be truncated if number of  line > max_lines. Check if file will be truncated every 1000 new lines.
- [x] configurable log file location. Save the config in the ~/.cortexrc 
```
logFolderPath: C:\Users\jan\cortexcpp
dataFolderPath: C:\Users\jan\cortexcpp
host: 127.0.0.1
port: 3928
```

dan-menlo:
@nguyenhoangthuan99 Please link this PR to issue via keyword - I've helped to do this in the Description

dan-menlo:
@nguyenhoangthuan99 A few questions: 

- Do we have a filesize estimate of how much space 100,000 log lines will equate to? 
- Should we make num log lines a variable in `.cortexrc` that users can configure? (need to find a good variable name for this)

I imagine cortex.cpp may have some embedded use-cases in the future, where devs may have diskspace constraints for logs (difficult for me to imagine now with >1gb model sizes, but this may change in the future)

engine/utils/cortex_utils.h
```diff
@@ -34,8 +34,10 @@ constexpr static auto kTensorrtLlmPath = "/engines/cortex.tensorrt-llm";
 
 inline std::string models_folder = "./models";
 inline std::string logs_folder = "./logs";
-inline std::string logs_base_name = "./logs/cortex";
+inline std::string logs_base_name = "./logs/cortex.log";
+inline std::string logs_cli_base_name = "./logs/cortex-cli.log";
 inline size_t log_file_size_limit = 20000000;  // ~20 mb
+inline size_t log_file_max_lines = 100000;     // 100,000 lines
```
dan-menlo: @nguyenhoangthuan99 I think we should have this as a variable from `.cortexrc`

dan-menlo (changes requested):

engine/utils/cortex_utils.h
```diff
@@ -34,8 +34,10 @@ constexpr static auto kTensorrtLlmPath = "/engines/cortex.tensorrt-llm";
 
 inline std::string models_folder = "./models";
 inline std::string logs_folder = "./logs";
-inline std::string logs_base_name = "./logs/cortex";
+inline std::string logs_base_name = "./logs/cortex.log";
+inline std::string logs_cli_base_name = "./logs/cortex-cli.log";
 inline size_t log_file_size_limit = 20000000;  // ~20 mb
```
dan-menlo: @nguyenhoangthuan99 Is this variable still used?

dan-menlo (changes requested):

engine/utils/cortex_utils.h
nguyenhoangthuan99: for now, we don't use the file size limit anymore. I will update it

engine/utils/cortex_utils.h
nguyenhoangthuan99: I'll move `log_file_max_lines` to read from .cortexrc

nguyenhoangthuan99:
Hi @dan-homebrew , Average each line will contains < 300 characters , each character is 1 byte. So 100,000 lines will be 300 * 100,000/(1024*1024) ~ 30Mb

dan-menlo:
> Hi @dan-homebrew , Average each line will contains < 300 characters , each character is 1 byte. So 100,000 lines will be 300 * 100,000/(1024*1024) ~ 30Mb

This looks good to me - let's include it as a default param in `.cortexrc`

engine/main.cc
```diff
@@ -29,16 +30,18 @@ void RunServer() {
   LOG_INFO << "Host: " << config.apiServerHost << " Port: " << config.apiServerPort << "\n";
 
   // Create logs/ folder and setup log to file
-  std::filesystem::create_directory(cortex_utils::logs_folder);
-  trantor::AsyncFileLogger asyncFileLogger;
-  asyncFileLogger.setFileName(cortex_utils::logs_base_name);
+  std::filesystem::create_directory(config.logFolderPath + "/" +
+                                    cortex_utils::logs_folder);
+  trantor::FileLogger asyncFileLogger;
+  asyncFileLogger.setFileName(config.logFolderPath + "/" +
```
namchuai, line 36: not sure if `/` works in windows. If you tested and it's work then please ignore this comment.

nguyenhoangthuan99: I tested in windows and it can works._

engine/utils/file_logger.cc
```diff
@@ -0,0 +1,175 @@
...
+  OpenFile();
+  LoadExistingLines();
+  TruncateFileIfNeeded();
+}
+
+FileLogger::CircularLogFile::~CircularLogFile() {
+  CloseFile();
+}
+void FileLogger::CircularLogFile::writeLog(const char* logLine,
+                                           const uint64_t len) {
```
nguyenhoangthuan99, line 42: need to add mutex here for thread safe

engine/utils/config_yaml_utils.h
```diff
@@ -48,8 +53,16 @@ inline CortexConfig FromYaml(const std::string& path,
 
   try {
     auto node = YAML::LoadFile(config_file_path.string());
+    int max_lines_;
```
vansangpfiev: naming convention: should rename to `max_lines`

engine/utils/file_logger.h
```diff
@@ -0,0 +1,68 @@
...
+
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+namespace trantor {
+
+class TRANTOR_EXPORT FileLogger : public AsyncFileLogger {
```
vansangpfiev, line 19: Please mark function with `override` keyword if implements virtual function inherit from `AsyncFileLogger`

dan-menlo (approved):
Lgtm

engine/utils/file_logger.h
```diff
@@ -0,0 +1,68 @@
...
+    ~CircularLogFile();
+
+    void writeLog(const char* logLine, const uint64_t len);
+    void flush();
+    uint64_t getLength() const { return totalLines_; }
+
+   private:
+    FILE* fp_{nullptr};
+    uint64_t max_lines_;
+    uint64_t totalLines_{0};
```
vansangpfiev: I'm not sure why we need `totalLines_` variable. Can you please elaborate?

nguyenhoangthuan99: Seem like we don't need it, I removed `totalLines_`

vansangpfiev (approved):

dan-menlo:
Great job!

PR #1179, opened 2024-09-09
adding libcurl to support pause/resume download

namchuai:
## Describe Your Changes

- Refactor, remove post processing download (extract, parse yaml, etc.) from download service.
- Adding percentage download (default from `libcurl`)

## Remaining things to do
- [ ] Haven't tested with `cortex update`

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/engine/commands/cortex_upd_cmd.h
+++ b/engine/commands/cortex_upd_cmd.h
@@ -1,5 +1,4 @@
 #pragma once
-#include <optional>
 #include <string>
 
 #include "httplib.h"
@@ -11,7 +10,7 @@ namespace commands {
 #ifndef CORTEX_VARIANT
 #define CORTEX_VARIANT file_manager_utils::kProdVariant
 #endif
-constexpr const auto kNightlyHost = "https://delta.jan.ai";
+constexpr const auto kNightlyHost = "delta.jan.ai";
 constexpr const auto kNightlyFileName = "cortex-nightly.tar.gz";
 const std::string kCortexBinary = "cortex";
 
@@ -113,12 +112,10 @@ inline bool ReplaceBinaryInflight(const std::filesystem::path& src,
 
 class CortexUpdCmd {
  public:
-  CortexUpdCmd();
   void Exec(std::string version);
 
  private:
   bool GetStableAndBeta(const std::string& v);
   bool GetNightly(const std::string& v);
 };
-
 }  // namespace commands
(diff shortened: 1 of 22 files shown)


engine/utils/url_parser.h
```diff
@@ -0,0 +1,84 @@
+#include <regex>
+#include <sstream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+#include "exceptions/malformed_url_exception.h"
+
+namespace url_parser {
```
namchuai, line 9: self note: write unit test

engine/commands/cortex_upd_cmd.h
```diff
@@ -11,7 +10,7 @@ namespace commands {
 #ifndef CORTEX_VARIANT
 #define CORTEX_VARIANT file_manager_utils::kProdVariant
 #endif
-constexpr const auto kNightlyHost = "https://delta.jan.ai";
+constexpr const auto kNightlyHost = "delta.jan.ai";
```
dan-menlo, line 13: Should these be pulled out into a `.cortexrc`? It seems very fragile to hard code a lot of these? cc @vansangpfiev @nguyenhoangthuan99

engine/utils/cortexso_parser.h
```diff
@@ -6,6 +6,7 @@
 #include <services/download_service.h>
 #include <nlohmann/json.hpp>
 #include "httplib.h"
+#include "utils/file_manager_utils.h"
 
 namespace cortexso_parser {
```
dan-menlo, line 11: What are we trying to refer to by naming this `cortexso_parser`? 
We might need a better name for this - i.e. this refers to a Cortex Model Repo (i.e. Git repo, tag-based)

dan-menlo (commented):
@namchuai I will let @vansangpfiev and @nguyenhoangthuan99 review this, but left a couple of comments on naming and architecture.

vansangpfiev (approved):

engine/utils/cortexso_parser.h
namchuai: I agree. But I will KIV this. Ideally, we will have multiple url parsers.

engine/commands/cortex_upd_cmd.h
namchuai: `.cortexrc` should be for things that are configurable for user? I doubt it's a suitable place.
And, IMO, we are not hard-coded here. We are defining compile time constants.

PR #1200, opened 2024-09-12
Fix: create log folder before logging

### LINKED ISSUE #1198: bug: `cortex-nightly` unable to run on first install
### Cortex version

v1.0.0

### Describe the Bug

`cortex-nightly` seems to be passed an incorrect log path: 

### Steps to Reproduce

_No response_

### Screenshots / Logs

image

### What is your OS?

- [X] MacOS
- [ ] Windows
- [ ] Linux

- [ ] cortex.llamacpp (default)

nguyenhoangthuan99:
Fix #1198 
Summary changes:
- Add end to end test for start server
- Add end to end test for create log folder after server started

### DIFF
--- a/engine/e2e-test/test_create_log_folder.py
+++ b/engine/e2e-test/test_create_log_folder.py
@@ -1,0 +1,24 @@
+import platform
+import os
+from pathlib import Path
+import pytest, requests
+from test_runner import run
+from test_runner import start_server, stop_server
+
+
+class TestCreateLogFolder:
+    @pytest.fixture(autouse=True)
+    def setup_and_teardown(self):
+        # Setup
+        success = start_server()
+        if not success:
+            raise Exception("Failed to start server")
+
+        yield
+
+        # Teardown
+        stop_server()
+
+    def test_create_log_folder_run_successfully(self):
+        root = Path.home()
+        assert os.path.exists(root / "cortexcpp" / "logs")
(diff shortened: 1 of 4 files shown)


engine/main.cc
```diff
@@ -31,6 +31,7 @@ void RunServer() {
            << " Port: " << config.apiServerPort << "\n";
 
   // Create logs/ folder and setup log to file
+  std::filesystem::create_directory(config.logFolderPath);
```
namchuai: using `std::filesystem::create_directories` should be better. Seems like it handle create directory recursively.
https://en.cppreference.com/w/cpp/filesystem/create_directory

Also, I'm worry about concatenate path using `/` might not work on windows. Should using `std::filesystem::path`

dan-menlo (approved):
@nguyenhoangthuan99 Approved to unblock. However: 

- Please add tests! 
- This is the sort of thing that an intern will not realize, remove folder creation, and end up bricking the entire product

engine/e2e-test/test_create_log_folder.py
```diff
@@ -0,0 +1,24 @@
...
+from pathlib import Path
+import pytest, requests
+from test_runner import run
+from test_runner import start_server, stop_server
+
+
+class TestCreateLogFolder:
+    @pytest.fixture(autouse=True)
+    def setup_and_teardown(self):
+        # Setup
```
namchuai, line 12: should remove the folder before start server to be sure that cortex automatically create mandatory folder

namchuai (approved):

vansangpfiev (approved):

PR #1237, opened 2024-09-17
feat: download pre-release engine

### LINKED ISSUE #1230: `cortex` should support pull a pre-release engine
We can only pull the latest engine for now. It would be nice if we can pull a pre-release build so we can test it before bump the latest.

vansangpfiev:
## Describe Your Changes

Syntax:
```
cortex engines install <engine> -v <version>
```

## Fixes Issues

- Closes https://github.com/janhq/cortex.cpp/issues/1230

## Self Checklist


engine/main.cc
```diff
@@ -206,12 +206,13 @@ int main(int argc, char* argv[]) {
           verbose = true;
         }
       }
+
+      trantor::FileLogger asyncFileLogger;
```
vansangpfiev, line 210: Fixes bug for `asyncFileLogger` out of scope

engine/services/engine_service.cc
```diff
@@ -76,16 +76,42 @@ std::vector<EngineInfo> EngineService::GetEngineInfoList() const {
...
-      .pathParams = {"repos", "janhq", engine, "releases", version},
+      .pathParams = get_params(),
   };
 
   httplib::Client cli(url_obj.GetProtocolAndHost());
   if (auto res = cli.Get(url_obj.GetPathAndQuery());
       res->status == httplib::StatusCode::OK_200) {
     auto body = json::parse(res->body);
+    auto get_data =
+        [&version](const nlohmann::json& json_data) -> nlohmann::json {
```
namchuai, line 98: Just a question. Is `std::optional<nlohmann::json>` is better? Can just return `std::nullopt` if not found instead of object `nlohmann::json()`

namchuai (approved):
✅ Code looks good to me. Thanks for the test!

engine/services/engine_service.cc
vansangpfiev: We assign new value to `body` if version is not `latest`, so I think it does not make sense to use std::optional in this case.

PR #1246, opened 2024-09-18
Feat/gguf yaml parser

### LINKED ISSUE #1244: Update GGUF parser and yaml parser for new model.yml
Related epic #1154 

We need to update parsers to generate `model.yml` template like below format
```
# BEGIN GENERAL GGUF METADATA
model: gemma-2-9b-it-Q8_0 # Model ID which is used for request construct - should be unique between models (author / quantization)
name: Llama 3.1      # metadata.general.name
version: 1           # metadata.version
sources:             # can be universal protocol (models://) OR absolute local file path (file://) OR https remote URL (https://)
  - models://huggingface/bartowski/Mixtral-8x22B-v0.1/main/Mixtral-8x22B-v0.1-IQ3_M-00001-of-00005.gguf # for downloaded model from HF
  - files://C:/Users/user/Downloads/Mixtral-8x22B-v0.1-IQ3_M-00001-of-00005.gguf # for imported model
# END GENERAL GGUF METADATA

# BEGIN INFERENCE PARAMETERS
# BEGIN REQUIRED
stop:                # tokenizer.ggml.eos_token_id
  - <|end_of_text|>
  - <|eot_id|>
  - <|eom_id|>
# END REQUIRED
# BEGIN OPTIONAL
stream: true         # Default true?
top_p: 0.9           # Ranges: 0 to 1
temperature: 0.6     # Ranges: 0 to 1
frequency_penalty: 0 # Ranges: 0 to 1
presence_penalty: 0  # Ranges: 0 to 1
max_tokens: 8192     # Should be default to context length
# END OPTIONAL
# END INFERENCE PARAMETERS

# BEGIN MODEL LOAD PARAMETERS
# BEGIN REQUIRED
prompt_template: |+  # tokenizer.chat_template
  <|begin_of_text|><|start_header_id|>system<|end_header_id|>

  {system_message}<|eot_id|><|start_header_id|>user<|end_header_id|>

  {prompt}<|eot_id|><|start_header_id|>assistant<|end_header_id|>
# END REQUIRED
# BEGIN OPTIONAL
ctx_len: 0          # llama.context_length | 0 or undefined = loaded from model
ngl: 33             # Undefined = loaded from model
# END OPTIONAL
# END MODEL LOAD PARAMETERS
```

nguyenhoangthuan99:
Fix #1244 

This is result of tinyllama.yaml file after downloaded successfully from cortexso.
image

### DIFF
--- a/engine/test/components/test_gguf_parser.cc
+++ b/engine/test/components/test_gguf_parser.cc
@@ -1,0 +1,160 @@
+#include "gtest/gtest.h"
+#include "config/gguf_parser.h"
+#include "config/yaml_config.h"
+#include <fstream>
+#include <cstdio>
+#include <vector>
+#include <stdexcept>
+#include <iostream>
+#include <iomanip>
+
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+class GGUFParserTest : public ::testing::Test {
+protected:
+    void SetUp() override {
+        gguf_handler = new config::GGUFHandler();
+        yaml_handler = new config::YamlHandler();
+    }
+
+    void TearDown() override {
+        delete gguf_handler;
+        delete yaml_handler;
+    }
+
+    config::GGUFHandler* gguf_handler;
+    config::YamlHandler* yaml_handler;
+
+    std::string getTempFilePath(const std::string& prefix, const std::string& extension) {
+    #ifdef _WIN32
+        char temp_path[MAX_PATH];
+        char file_name[MAX_PATH];
+        GetTempPathA(MAX_PATH, temp_path);
+        GetTempFileNameA(temp_path, prefix.c_str(), 0, file_name);
+        std::string path(file_name);
+        DeleteFileA(file_name);  // Delete the file created by GetTempFileNameA
+        return path + extension;
+    #else
+        std::string path = "/tmp/" + prefix + "XXXXXX" + extension;
+        char* temp = strdup(path.c_str());
+        int fd = mkstemps(temp, extension.length());
+        if (fd == -1) {
+            free(temp);
+            throw std::runtime_error("Failed to create temporary file");
+        }
+        close(fd);
+        std::string result(temp);
+        free(temp);
+        return result;
+    #endif
+    }
+
+    std::string createMockGGUFFile() {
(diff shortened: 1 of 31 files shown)


dan-menlo:
@nguyenhoangthuan99 Can I just check: 

- I think from our earlier discussion, `model.yaml` is optional (we can load params directly from gguf)
- However, is this done by us auto-creating a `model.yaml` for every GGUF file (e.g. upon `pull` or `run`)

The reason why I'm asking is: 
- Do we need a model method to create a `model.yaml` by parsing from a GGUF file?

nguyenhoangthuan99:
I think we still need it now because:
-  There is 1 field `chat_template` we still need to parse from gguf file and render chat template. The implementation of `cortex.llamacpp` requires rendered chat template. 
- Stop tokens also needed to parse from gguf because `cortex.llamacpp` requires that field when run chat completion.
- For further feature like model-compatibility API, we need the `llama.embedding_length`, `llama.attention.head_count`, `llama.attention.head_count_kv` to calculate the memory constraint for kv cache and recommend suitable `ctx_len` for user
- `ngl` <> `llama.block_count` is also can be read inside gguf file and can be used to recommend the suitable `ngl` to user's hardware

nguyenhoangthuan99:
I'll add uni-tests for this PR

dan-menlo:
Got it. So we will still create a model.yaml whenever we pull, run or import a model?

nguyenhoangthuan99:
yes, we need to create model.yaml with the following cases:
- pull model from other sources
- import a random model from local

Models from cortexso already have model.yml so we don't need to create `model.yml`, just update the file location of binary inside the `model.yml`

dan-menlo (approved):
Approving to unblock. Will let @vansangpfiev review for code quality.

vansangpfiev (approved):

engine/test/components/test_gguf_parser.cc
```diff
@@ -0,0 +1,160 @@
...
+#ifdef _WIN32
+#include <windows.h>
+#else
+#include <unistd.h>
+#endif
+
+class GGUFParserTest : public ::testing::Test {
+protected:
+    void SetUp() override {
+        gguf_handler = new config::GGUFHandler();
```
vansangpfiev: Can use std::unique_ptr instead of raw pointer

PR #1256, opened 2024-09-19
model set alias command

### LINKED ISSUE #1255: Model alias command
# Add Model Alias Command

related epic #1154 

## Feature Overview
This PR introduces a new command to update the alias of an existing model in our CLI tool. The model alias feature allows users to assign and update custom, memorable names for their models, enhancing usability and model management.

## Implementation Details
- Extended `ModelListUtils` with an `UpdateModelAlias` method
- Updated model list persistence to include and manage aliases
- Only update new alias if it is unique, and the model to be updated is existed

### Key Functions
- `ModelAliasCmd::Exec(const std::string& model_handle, const std::string& model_alias)`
- `ModelListUtils::UpdateModelAlias(const std::string& model_id, const std::string& new_model_alias)`

## Usage
Users can update a model's alias using the following command:
```
cli models alias --model_id <model_id_or_current_alias> --new_model_alias <new_alias>
```

## Testing
- Added unit tests for `ModelListUtils::UpdateModelAlias` in `model_list_utils_test.cpp`
- Manual testing performed across Windows, macOS, and Linux platforms

nguyenhoangthuan99:
Implementation of model alias command
fix #1255

### DIFF
--- a/engine/controllers/command_line_parser.cc
+++ b/engine/controllers/command_line_parser.cc
@@ -156,7 +156,7 @@ bool CommandLineParser::SetupCommand(int argc, char** argv) {
       models_cmd->add_subcommand("alias", "Add alias name for a modelID");
   model_alias_cmd->add_option("--model_id", model_id, "Can be modelID or model alias to identify model");
   model_alias_cmd->require_option();
-  model_alias_cmd->add_option("--new_model_alias", model_alias, "new alias to be set");
+  model_alias_cmd->add_option("--alias", model_alias, "new alias to be set");
   model_alias_cmd->require_option();
   model_alias_cmd->callback([&model_id, &model_alias]() {
     commands::ModelAliasCmd mdc;
--- a/engine/controllers/command_line_parser.cc
+++ b/engine/controllers/command_line_parser.cc
@@ -156,9 +156,9 @@ bool CommandLineParser::SetupCommand(int argc, char** argv) {
   std::string model_alias;
   auto model_alias_cmd =
       models_cmd->add_subcommand("alias", "Add alias name for a modelID");
-  model_alias_cmd->add_option("--model_id", model_id, "");
+  model_alias_cmd->add_option("--model_id", model_id, "Can be modelID or model alias to identify model");
   model_alias_cmd->require_option();
-  model_alias_cmd->add_option("--model_alias", model_alias, "");
+  model_alias_cmd->add_option("--new_model_alias", model_alias, "new alias to be set");
   model_alias_cmd->require_option();
   model_alias_cmd->callback([&model_id, &model_alias]() {
     commands::ModelAliasCmd mdc;
--- a/engine/controllers/command_line_parser.cc
+++ b/engine/controllers/command_line_parser.cc
@@ -6,6 +6,7 @@
 #include "commands/engine_install_cmd.h"
 #include "commands/engine_list_cmd.h"
 #include "commands/engine_uninstall_cmd.h"
+#include "commands/model_alias_cmd.h"
 #include "commands/model_del_cmd.h"
 #include "commands/model_get_cmd.h"
 #include "commands/model_list_cmd.h"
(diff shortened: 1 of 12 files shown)


engine/e2e-test/test_cli_engine_uninstall.py
```diff
@@ -18,7 +18,7 @@ def setup_and_teardown(self):
 
     def test_engines_uninstall_llamacpp_should_be_successfully(self):
         exit_code, output, error = run(
-            "Uninstall engine", ["engines", "uninstall", "cortex.llamacpp"]
+            "Uninstall engine", ["engines", "uninstall", "cortex.llamacpp"], timeout = None
```
namchuai: Uninstall should be quick. I don't think we should put None here. Are you experiencing a slow when uninstall engine?

nguyenhoangthuan99: Yes, I should remove timeout for uninstall engine.

engine/controllers/command_line_parser.cc
```diff
@@ -150,6 +151,18 @@ bool CommandLineParser::SetupCommand(int argc, char** argv) {
     mdc.Exec(model_id);
   });
 
+  std::string model_alias;
+  auto model_alias_cmd =
+      models_cmd->add_subcommand("alias", "Add alias name for a modelID");
+  model_alias_cmd->add_option("--model_id", model_id, "Can be modelID or model alias to identify model");
+  model_alias_cmd->require_option();
+  model_alias_cmd->add_option("--new_model_alias", model_alias, "new alias to be set");
```
dan-menlo: @nguyenhoangthuan99 Small nitpick: I recommend we simplify this to just `--alias`

engine/controllers/models.cc
```diff
@@ -192,4 +192,56 @@ void Models::DeleteModel(const HttpRequestPtr& req,
...
+      ret["result"] = "OK";
+      ret["modelHandle"] = model_handle;
+      ret["message"] = message;
+      auto resp = cortex_utils::CreateCortexHttpJsonResponse(ret);
+      resp->setStatusCode(k200OK);
+      callback(resp);
+    } else {
+      std::string message = "Unable to set model alias for modelID '" +
+                            model_handle + "': model alias '" + model_alias +
+                            "' or modelID is not unique!";
```
dan-menlo: Nitpick: This scenario would only occur if model alias is already used right? 

- User might be confused by an error message suggesting `model_id` is not unique
- As an aside, how do we plan to deal with non-unique `model_id`s?

engine/e2e-test/test_cli_engine_install.py
```diff
@@ -8,7 +8,7 @@ class TestCliEngineInstall:
 
     def test_engines_install_llamacpp_should_be_successfully(self):
         exit_code, output, error = run(
-            "Install Engine", ["engines", "install", "cortex.llamacpp"], timeout=60
+            "Install Engine", ["engines", "install", "cortex.llamacpp"], timeout=None
```
dan-menlo: Woah - this seems a bit dangerous, I would prefer to have a timeout bound (you can set it at 10x magnitude)

dan-menlo (approved):
Approved with small nitpicks

engine/controllers/command_line_parser.cc
nguyenhoangthuan99: I'll update it

engine/controllers/models.cc
nguyenhoangthuan99: - Yes, let me modify the error message, this scenario occurs when model alias is already in use or equal to model_id of other models.
- I think the model-id should be unique to make model entry can be distinguished, alias just a shorter name for easier to use.

engine/e2e-test/test_cli_engine_install.py
nguyenhoangthuan99: I'll update it

vansangpfiev (approved):

namchuai (approved):
✅ lgtm

PR #1271, opened 2024-09-19
Feat/model get

### LINKED ISSUE #1075: feat: `cortex models get <MODEL_ID>`  and `cortex models list`

nguyenhoangthuan99:
Implement `list` and `get` command

Fix #1075 

Logic flow:
- Read entries in `model.list`
- Get information about `model_id` and `path/to/model.yml`
- Read yaml file and return information

### Model list example output
image

### Model get example output
image


namchuai:
I think we should move the print logic of `ModelConfig` to its struct member function (ToString() for example) to reduce the code duplication between CLI and controller.

IMO, Model's logic stuffs should be placed in `ModelService`. In the future, we will need to centralize this ModelService so that we can easily get what model is running (for example). CLI and Http API should just serve as an interface to communicate.

namchuai (approved):

namchuai:
✅ lgtm

vansangpfiev (approved):

PR #1277, opened 2024-09-20
fix: should check model status before start it

vansangpfiev:
## Describe Your Changes

llamacpp engine does not handle the case when we load a model second time. Should check the model's status before starting it.

## Fixes Issues

- https://github.com/janhq/cortex.cpp/issues/1272

## Self Checklist

### DIFF
--- a/engine/controllers/server.cc
+++ b/engine/controllers/server.cc
@@ -342,20 +342,22 @@ void server::LoadModel(const HttpRequestPtr& req,
     auto func =
         engines_[engine_type].dl->get_function<EngineI*()>("get_engine");
     engines_[engine_type].engine = func();
+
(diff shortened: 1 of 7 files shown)


engine/controllers/server.cc
```diff
@@ -342,20 +342,22 @@ void server::LoadModel(const HttpRequestPtr& req,
     auto func =
         engines_[engine_type].dl->get_function<EngineI*()>("get_engine");
     engines_[engine_type].engine = func();
+
```
vansangpfiev, line 345: Only set logger for engine once, or it will crash if load/unload/load model

namchuai (approved):

nguyenhoangthuan99 (approved):

PR #1309, opened 2024-09-23
Model update command/api

### LINKED ISSUE #1154: epic: Implement new Model Folder and model.yaml
## Goal

- We should have a model folder that is able to handle different models
    - Built-in models (e.g. `janhq/llama3:7b-tensorrt-llm`)
    - Huggingface GGUF repos with multiple quants (e.g. `bartowski/llama3-gguf`)
    - Huggingface specific GGUF (may have multiple from same directory)
    - In future: Nvidia NGC or TensorRT Cloud
- Do we use sub-folders?
- How does `model.yaml` work?
- Model detection should not depend on model folder

## Tasklist

- [X] #1300 
- [x] #1320 
- [x] https://github.com/janhq/cortex.cpp/issues/1121
- [x] https://github.com/janhq/cortex.cpp/issues/1241

## Decisions

- https://github.com/janhq/cortex.cpp/discussions/1113
- https://github.com/janhq/cortex.cpp/discussions/1123
- #1178
- Legacy model folder structure: https://github.com/janhq/jan/issues/3541#issuecomment-2328413473

## Bugs
- [x] https://github.com/janhq/cortex.cpp/issues/1270
- [x] https://github.com/janhq/cortex.cpp/issues/1274

## Edge Cases
- What if we download multiple GGUFs from the same Huggingface repo? 
    - Saved in the same repo folder, with multiple .gguf files (see https://github.com/janhq/cortex.cpp/issues/1320#issuecomment-2387671325)
- How does `cortex model update <model>` work?
    - https://github.com/janhq/cortex.cpp/issues/1121

nguyenhoangthuan99:
Closes #1154 

command option:
image
Now cortex.cpp can update model parameters by passing it to this command, only passed argument will be updated

If successfully Update, it will output which parameters of model ID is updated

image


dan-menlo:
@nguyenhoangthuan99 Can I check the main purpose of the `cortex models update` is for Jan to be able update model parameters in `model.yaml`? 

- In the future, we will need clear differentiation by model engine, as the parameters differ by each engine

engine/controllers/command_line_parser.cc
```diff
@@ -457,3 +456,104 @@ void CommandLineParser::EngineGet(CLI::App* parent) {
...
+
+void CommandLineParser::ModelUpdate(CLI::App* parent) {
+  auto model_update_cmd =
+      parent->add_subcommand("update", "Update configuration of a model");
+  model_update_cmd->group(kSubcommands);
+  model_update_cmd->add_option("--model_id", cml_data_.model_id, "Model ID")
+      ->required();
+
+  // Add options dynamically
+  model_update_cmd->add_option("--name", model_update_options_.name, "Name");
```
dan-menlo: @vansangpfiev @namchuai @nguyenhoangthuan99 I know this is C++, but I wonder whether there is a way to automatically generate this sort of repetitive code from an array of param names. 

- `command_line_parser.cc`
- `model_config.h`
- `model_upd_cmd.h`

That said, with AI, I think repetitive coding can be handled via LLM.

dan-menlo (approved):

engine/controllers/command_line_parser.cc
freelerobot: Suggestion: pass it through Jan/endpoint with a refactor prompt 🙏

nguyenhoangthuan99: Got it, I refactored it to make it more clear

engine/controllers/command_line_parser.cc
```diff
@@ -457,3 +462,36 @@ void CommandLineParser::EngineGet(CLI::App* parent) {
         [engine_name] { commands::EngineGetCmd().Exec(engine_name); });
   }
 }
+
+void CommandLineParser::ModelUpdate(CLI::App* parent) {
+  auto model_update_cmd =
+      parent->add_subcommand("update", "Update configuration of a model");
+  model_update_cmd->group(kSubcommands);
```
namchuai, line 454: please reformat the code

engine/controllers/command_line_parser.cc
```diff
@@ -457,3 +462,36 @@ void CommandLineParser::EngineGet(CLI::App* parent) {
...
+  model_update_cmd->add_option("--model_id", cml_data_.model_id, "Model ID")
+      ->required();
+
+  // Add options dynamically
```
namchuai, line 458: may be can remove this comment?

engine/commands/model_upd_cmd.cc
```diff
@@ -0,0 +1,127 @@
...
+    CLI_LOG("Successfully updated model ID '" + model_handle_ + "'!");
+  } catch (const std::exception& e) {
+    CLI_LOG("Failed to update model with model ID '" + model_handle_ +
+            "': " + e.what());
+  }
+}
+
+void ModelUpdCmd::UpdateConfig(const std::string& key,
+                               const std::string& value) {
+  static const std::unordered_map<
```
namchuai, line 34: Hmm, I think there's a better way. We can define a function: UpdateModelConfig(const std::string& attr, std::variant<std::string, explicit_int, explicit_bool>) and handle the update logic there.

However if this works for you then I'm fine with it.

namchuai (approved):

engine/commands/model_upd_cmd.cc
vansangpfiev: I think should keep it this way. We don't know which format user will input, so need to handle it for specific cases

vansangpfiev (approved):

PR #1334, opened 2024-09-25
fix: save latest release info

vansangpfiev:
## Describe Your Changes
- Add `latestRelease` and `checkedForUpdateAt` to the `.cortexrc`
- Interval to check for new update: nightly 10 mins, beta 1 hour, stable 24 hours
- Move some long functions to .cc file to save build time

Examples:
If current version is the latest:
```
> .\cortex-nightly.exe update
cortex is up to date
```
## Fixes Issues

- https://github.com/janhq/cortex.cpp/issues/1278
- https://github.com/janhq/cortex.cpp/issues/1268

## Self Checklist


engine/commands/cortex_upd_cmd.cc
```diff
@@ -12,7 +12,182 @@
...
+  } catch (const std::exception& e) {
+    CTL_ERR("Something went wrong: " << e.what());
+    restore_binary();
+    return false;
+  }
+
+  return true;
+}
+
 void CortexUpdCmd::Exec(std::string v) {
```
namchuai: just a question, why not `const std::string& v`?

vansangpfiev: Nice catch!  Seems like we intended to use std::move() in early day. Will fix it

namchuai (approved):
✅ thanks for the tests

PR #1352, opened 2024-09-27
fix: work around for connection hang when network down

### LINKED ISSUE #1350: bug: check new update hang if network is down
### Cortex version

v0.5.0-120

### Describe the Bug

Check new update for will be hanged if network is down

### Steps to Reproduce

run `cortex -v`

### Screenshots / Logs

_No response_

### What is your OS?

- [ ] MacOS
- [X] Windows
- [X] Linux

- [X] cortex.llamacpp (default)

vansangpfiev:
## Describe Your Changes

When the network is down, our process is hang for 10 seconds to wait for HTTP response. The httplib does not support connection timeout in this case yet, so we need to work around.

## Fixes Issues

- Closes https://github.com/janhq/cortex.cpp/issues/1350

## Self Checklist

### DIFF
--- a/engine/controllers/command_line_parser.cc
+++ b/engine/controllers/command_line_parser.cc
@@ -73,7 +73,7 @@ bool CommandLineParser::SetupCommand(int argc, char** argv) {
 #ifdef CORTEX_CPP_VERSION
   if (cml_data_.check_upd) {
     // TODO(sang) find a better way to handle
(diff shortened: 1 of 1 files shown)


dan-menlo:
Thanks @hiento09 for flagging this up!

namchuai (approved):
✅ lgtm. Having another thread is ok, but quite expensive. I think we can try with `libcurl` in the future.

PR #1379, opened 2024-10-01
fix: correct model_id in chat

### LINKED ISSUE #1372: bug: No engine field in request body
### Problem Statement

I'm sending a chat/completions request to the cortex.cpp endpoints, setting up my daily use tools. The request is fully OpenAI compatible. I always receive an error response similar to this:

```json
{
    "message": "No engine field in request body"
}
```

### Feature Idea

Since the cortex.cpp manages to retrieve model information, it should select the engine accordingly. This eliminates the need for an extra `engine` field in the request, ensuring that 100% OpenAI-compatible products can utilize the endpoint.

vansangpfiev:
## Describe Your Changes

- Add `/models/start` and `/models/stop` endpoint
- Fallback to cortex.llamacpp if engine is not in request body
- Use model_handle for model chat id

## Fixes Issues

- fixes https://github.com/janhq/cortex.cpp/issues/1372

## Self Checklist

### DIFF
--- a/engine/commands/chat_completion_cmd.h
+++ b/engine/commands/chat_completion_cmd.h
@@ -9,8 +9,8 @@ class ChatCompletionCmd {
  public:
   void Exec(const std::string& host, int port, const std::string& model_handle,
             std::string msg);
-  void Exec(const std::string& host, int port, const config::ModelConfig& mc,
-            std::string msg);
+  void Exec(const std::string& host, int port, const std::string& model_handle,
+            const config::ModelConfig& mc, std::string msg);
 
  private:
   std::vector<nlohmann::json> histories_;
--- a/engine/commands/chat_completion_cmd.cc
+++ b/engine/commands/chat_completion_cmd.cc
@@ -4,10 +4,11 @@
 #include "cortex_upd_cmd.h"
 #include "database/models.h"
 #include "model_status_cmd.h"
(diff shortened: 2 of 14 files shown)


engine/commands/chat_completion_cmd.h
```diff
@@ -9,8 +9,8 @@ class ChatCompletionCmd {
  public:
   void Exec(const std::string& host, int port, const std::string& model_handle,
             std::string msg);
-  void Exec(const std::string& host, int port, const config::ModelConfig& mc,
-            std::string msg);
+  void Exec(const std::string& host, int port, const std::string& model_handle,
+            const config::ModelConfig& mc, std::string msg);
```
namchuai, line 13: should be `const std::string& msg` ?

vansangpfiev: we use std::move, so it is not a problem. But I think should use `const std::string&` like other parameters. Will fix it shortly

namchuai (approved):
✅ lgtm

engine/commands/chat_completion_cmd.h
vansangpfiev: ah no, it is intended to check the empty condition here
```
std::string user_input = std::move(msg);
      if (user_input.empty()) {
```

PR #1403, opened 2024-10-03
Fix: redownload model.yml

nguyenhoangthuan99:
## Describe Your Changes

- Create default log path for engines at first time run cortexcpp
- Fix display issue when re download model.yml file. `model.yml` can be changed in local and file size can larger than original version in hugging face, this causes the problem that number of left bytes to download overflow.

## Fixes Issues
image


namchuai (approved):
✅ lgtm

vansangpfiev (approved):

PR #1495, opened 2024-10-16
Feat/support custom prompt template

### LINKED ISSUE #1484: feat: cortex support custom prompt_template when load model
This feature supports Jan or other applications that want to integrate Cortex as a server. When users want to change the prompt template, Cortex can directly load that prompt instead of saving it to model.yml and then loading the model from model.yml.

nguyenhoangthuan99:
## Describe Your Changes

- Allow custom prompt template when loadding model
image

## Fixes Issues

- Closes #1484


engine/services/model_service.cc
```diff
@@ -580,9 +581,19 @@ cpp::result<bool, std::string> ModelService::StartModel(
       return false;
     }
     json_data["model"] = model_handle;
-    json_data["system_prompt"] = mc.system_template;
-    json_data["user_prompt"] = mc.user_template;
-    json_data["ai_prompt"] = mc.ai_template;
+    if (custom_prompt_template.has_value() &&
+        !custom_prompt_template.value_or("").empty()) {
```
vansangpfiev: Only need to use (`value_or`) or (`has_value` and `value`)

engine/services/model_service.cc
```diff
@@ -580,9 +581,19 @@ cpp::result<bool, std::string> ModelService::StartModel(
...
+      auto& pt = custom_prompt_template.value();
+      json_data["system_prompt"] = pt.substr(0, pt.find_first_of('{'));
```
vansangpfiev: Can we add to string_utils? Seems like we can re-use this somewhere else.
Maybe return a struct with system, user, and ai prompt

vansangpfiev (approved):

PR #1503, opened 2024-10-16
Feat/function calling

### LINKED ISSUE #295: feat: Cortex supports Function Calling

nguyenhoangthuan99:
## Describe Your Changes

- support function calling in user message 

image

## Fixes Issues

- Closes #295


namchuai (approved):
lgtm

PR #1522, opened 2024-10-21
fix: update download api

namchuai:
## Describe Your Changes

- Allow user to specify their desired `model_id`.
- Add a thread safe data structure for DownloadTaskQueue. Working on refactor DownloadService to use new DownloadTaskQueue to support parallel download.
- Clean up alias API.

## Self Checklist


engine/cli/commands/model_list_cmd.cc
```diff
@@ -17,8 +17,9 @@ using namespace tabulate;
 using Row_t =
     std::vector<variant<std::string, const char*, string_view, Table>>;
 
-void ModelListCmd::Exec(const std::string& host, int port, std::string filter,
-                        bool display_engine, bool display_version) {
+void ModelListCmd::Exec(const std::string& host, int port,
+                        const std::string& filter, bool display_engine,
```
vansangpfiev, line 21: Nice! I missed this in the previous PR.

vansangpfiev (approved):

louis-jan:
This change broke Jan's cortex because it blocks custom engine path support. @vansangpfiev @namchuai

dan-menlo:
> This change broke Jan's cortex because it blocks custom engine path support. @vansangpfiev @namchuai

@louis-jan @vansangpfiev @namchuai Can I understand this a bit more in detail - what does it mean by blocking custom engine path support? 

Does this mean we have to revert?

louis-jan:
> > This change broke Jan's cortex because it blocks custom engine path support. @vansangpfiev @namchuai
> 
> @louis-jan @vansangpfiev @namchuai Can I understand this a bit more in detail - what does it mean by blocking custom engine path support?
> 
> Does this mean we have to revert?

Hey @dan-homebrew guys fixed it already. Applications embed Cortex, which bundles engines that aren't fixed at ~/cortex/engines since `engines install` won't be done in run-time. They could be `mac-arm/engines, mac-intel/engines, or windows/engines` or pre-packaged. And they can point to the corresponding engine at runtime via env or args.

PR #1524, opened 2024-10-21
fix: return model_id instead of error if model is already downloaded

vansangpfiev:
## Describe Your Changes

- Return model_id instead of error if model is already downloaded

## Fixes Issues

- https://github.com/janhq/cortex.cpp/issues/1513

## Self Checklist


namchuai (approved):
✅ lgtm

PR #1536, opened 2024-10-22
chore: refactor inference service

vansangpfiev:
## Describe Your Changes

- call inference service API instead of http request for start stop model

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


engine/services/inference_service.h
```diff
@@ -2,7 +2,9 @@
 
 #include <optional>
 #include <variant>
-#include "common/base.h"
+#include <mutex>
+#include <queue>
+#include <condition_variable>
```
namchuai: Do we use this import?

namchuai (approved):
✅ lgtm

engine/services/inference_service.h
vansangpfiev: No - we don't

engine/controllers/server.cc
```diff
@@ -95,15 +96,15 @@ void server::GetModels(const HttpRequestPtr& req,
...
+  auto ir = inference_svc_->GetEngines(req->getJsonObject());
   auto resp = cortex_utils::CreateCortexHttpJsonResponse(ir);
   callback(resp);
 }
 
 void server::FineTuning(
     const HttpRequestPtr& req,
     std::function<void(const HttpResponsePtr&)>&& callback) {
-  auto ir = inference_svc_.FineTuning(req->getJsonObject());
+  auto ir = inference_svc_->FineTuning(req->getJsonObject());
```
dan-menlo, line 107: @vansangpfiev Can I check: why do we have a FineTuning endpoint, if we haven't implemented it yet?

vansangpfiev: This is the legacy code from last time when we were trying to integrate `cortex.python` into `cortex.cpp`

PR #1542, opened 2024-10-23
fix: incorrect downloadedBytes callback

namchuai:
## Describe Your Changes

- Fix an issue that `downloadedBytes` is incorrect for each DownloadItem.

## Self Checklist

### DIFF
--- a/engine/services/download_service.h
+++ b/engine/services/download_service.h
@@ -80,8 +80,7 @@ class DownloadService {
  private:
   struct DownloadingData {
     std::string item_id;
-    DownloadTask* download_task;
-    EventQueue* event_queue;
+    DownloadService* download_service;
   };
 
   cpp::result<void, std::string> VerifyDownloadTask(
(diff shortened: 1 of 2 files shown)


engine/services/download_service.h
```diff
@@ -131,13 +132,23 @@ class DownloadService {
...
   static int ProgressCallback(void* ptr, curl_off_t dltotal, curl_off_t dlnow,
                               curl_off_t ultotal, curl_off_t ulnow) {
-    auto* downloading_data = static_cast<DownloadingData*>(ptr);
-    auto& event_queue = *downloading_data->event_queue;
-    auto& download_task = *downloading_data->download_task;
+    auto downloading_data = static_cast<DownloadingData*>(ptr);
+    if (downloading_data == nullptr) {
+      return 0;
+    }
+    auto dl_item_id = downloading_data->item_id;
```
vansangpfiev: Is there any problem if we use `auto const&` here?

vansangpfiev (approved):

engine/services/download_service.h
namchuai: can be const auto

PR #1601, opened 2024-11-01
fix: use download event type to listen ws on client side

vansangpfiev:
## Describe Your Changes

- We are using id to listen to ws event, it is fragile so should use download event type instead

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


namchuai (approved):
✅ lgtm

PR #1626, opened 2024-11-04
chore: return model size after pulled

louis-jan:
## Describe Your Changes

- Clients like Jan view the model's size as metadata for local models. This helps users determine compatibility with their computer specs.
- This PR saves the model's size as metadata (in bytes).

`model.yaml`
Screenshot 2024-11-04 at 14 54 03

`API Response body`
Screenshot 2024-11-04 at 14 58 06

API Playground
image

## Changes made

The provided git diff introduces changes primarily focused on tracking and handling a new attribute called `size` within the `ModelConfig` struct. Here are the key changes summarized:

1. **ModelConfig Struct Update**:
   - A new field `uint64_t size` is added to the `ModelConfig` struct to store the size of the model.
   - `FromJson` method is updated to parse the `size` attribute from JSON if present.
   - `ToJson` method is updated to include the `size` attribute in the JSON output.
   - `Print` method now prints the `size` attribute along with other attributes.

2. **YamlHandler Update**:
   - In `UpdateModelConfig`, the YAML node is updated to include the `size` of the model if it is not NaN. 
   - Likely missed the context, but casting `size` of type `uint64_t` to `double` for NaN check is unusual since `uint64_t` can't be NaN.

3. **ParseGguf Function Update**:
   - The function signature of `ParseGguf` is updated to include an optional `std::uint64_t size` parameter.
   - The `size` is set on `model_config` with a default of 0 if no `size` is provided.

4. **Model Service Update**:
   - In two functions (`HandleDownloadUrlAsync` and `HandleUrl`), after a download task is finished, the total downloaded size (`model_size`) is calculated by summing up `bytes` from all items in `finishedTask.items`.
   - `ParseGguf` is called with the computed `model_size` to ensure the size is recorded in the `ModelConfig`.

These changes enhance the system to manage and persist the size of model files across configurations and YAML updates.

### DIFF
--- a/engine/config/yaml_config.cc
+++ b/engine/config/yaml_config.cc
@@ -75,6 +75,8 @@ void YamlHandler::ModelConfigFromYaml() {
       tmp.model = yaml_node_["model"].as<std::string>();
     if (yaml_node_["version"])
       tmp.version = yaml_node_["version"].as<std::string>();
+    if (yaml_node_["size"])
+      tmp.size = yaml_node_["size"].as<uint64_t>();
     if (yaml_node_["engine"])
       tmp.engine = yaml_node_["engine"].as<std::string>();
     if (yaml_node_["prompt_template"]) {
@@ -266,6 +268,9 @@ void YamlHandler::UpdateModelConfig(ModelConfig new_model_config) {
     if (!model_config_.grammar.empty())
       yaml_node_["grammar"] = model_config_.grammar;
 
+    if (!std::isnan(static_cast<double>(model_config_.size)))
+      yaml_node_["size"] = model_config_.size;
+
     yaml_node_["created"] = std::time(nullptr);
   } catch (const std::exception& e) {
     std::cerr << "Error when update model config : " << e.what() << std::endl;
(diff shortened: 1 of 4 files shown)


vansangpfiev (approved):

engine/config/yaml_config.cc
```diff
@@ -266,6 +268,9 @@ void YamlHandler::UpdateModelConfig(ModelConfig new_model_config) {
     if (!model_config_.grammar.empty())
       yaml_node_["grammar"] = model_config_.grammar;
 
+    if (!std::isnan(static_cast<double>(model_config_.size)))
```
namchuai: Shouldn't this be `static_cast<uint64_t>`?

louis-jan: Nice catch. I reused all the checks above, even for int, and they all double cast check.

louis-jan: Also, uint64_t can never represent a nan value. Isn't it?

louis-jan: Removed the check as confirmed. I don't think it's necessary. Alex explained it's a CI hack.

PR #1630, opened 2024-11-04
chore: update model's size on import

louis-jan:
## Describe Your Changes

This PR updates the model's size on import as supported in the pull request. It includes minor refactoring of file size extension and consolidates features using the std::filesystem API.

## Changes made
The provided `git diff` reflects several modifications across multiple files in the project. Here's a concise summary of these changes:

1. **`gguf_parser.cc`**:
   - Replaced manual file size determination with `std::filesystem::file_size` for reliability and simplicity.
   - Removed redundant code related to the old method of fetching the file size.

2. **`models.cc`**:
   - Introduced `std::filesystem::file_size` to fetch the file size when importing models, storing this size in the `model_config`.

3. **`download_service.cc` and `download_service.h`**:
   - Removed the custom `GetLocalFileSize` method and replaced its usage with `std::filesystem::file_size` for determining file sizes.
   - The `download_service.h` file had the method deleted from class specifications.
   - Adjustments were made to ensure `curl_easy_setopt` uses the new size-fetching method directly.

These changes simplify and modernize file size handling by leveraging the C++17 `<filesystem>` library.

### DIFF
--- a/engine/controllers/models.cc
+++ b/engine/controllers/models.cc
@@ -352,11 +352,11 @@ void Models::ImportModel(
           modelPath, file_path,
           std::filesystem::copy_options::update_existing);
       model_config.files.push_back(file_path.string());
-      auto size = std::filesystem::file_size(file_path);
+      uintmax_t size = std::filesystem::file_size(file_path);
       model_config.size = size;
     } else {
       model_config.files.push_back(modelPath);
-      auto size = std::filesystem::file_size(modelPath);
+      uintmax_t size = std::filesystem::file_size(modelPath);
       model_config.size = size;
     }
     model_config.model = modelHandle;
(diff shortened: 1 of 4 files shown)


engine/controllers/models.cc
```diff
@@ -351,8 +352,12 @@ void Models::ImportModel(
           modelPath, file_path,
           std::filesystem::copy_options::update_existing);
       model_config.files.push_back(file_path.string());
+      uintmax_t size = std::filesystem::file_size(file_path);
+      model_config.size = size;
     } else {
       model_config.files.push_back(modelPath);
+      uintmax_t size = std::filesystem::file_size(modelPath);
```
namchuai: `auto size` might be better? we won't need the `#include <cstdint>`

louis-jan: Oh. I thought a clear type declaration would be better

namchuai (approved):
✅ thank you!

Issue #1638, opened 2024-11-05
planning: `cortex engines update` 


gabrielle-ong:
**Problem**
- I needed to upgrade cortex.llama-cpp from `0.1.37` to `0.1.37-01.11.24` to test our changes
- `cortex engines install llama-cpp` did not install the latest sub(?) version, it still installed v0.1.37
- This required manual `cortex engines install llama-cpp -v v0.1.37-01.11.24`. 
- I would have not known there was a newer version without looking at the releases page

Discussion:
1. Should we align the cortex.llama-cpp version names with [upstream llama.cpp](https://github.com/ggerganov/llama.cpp/releases) eg `0.1.37-b4033` instead of our date format `0.1.37 to 0.1.37-01.11.24` 
2. `cortex engines update llama-cpp`: we should have a way to update engines when new engines are available, and delete the old version?
3. Idea: `cortex update` also updates engines

**Success Criteria**
- 

**Tasklist**
- [ ]

namchuai:
> cortex engines install llama-cpp did not install the latest sub(?) version, it still installed v0.1.37

It's because we are using API to get latest version from Github and the latest version returned is v0.1.37.
```sh
curl -s "https://api.github.com/repos/janhq/cortex.llamacpp/releases/latest" | jq -r .tag_name
v0.1.37
```
So it will update the latest version after we mark a version as a stable release. IMHO, this is good for us. Since pre-release version can have some bug and we don't want to disturb the broad user base. We can promote a pre-release to stable when we feel confident to do so.

> This required manual cortex engines install llama-cpp -v v0.1.37-01.11.24.
I would have not known there was a newer version without looking at the releases page

You can use the below command to print a list of available versions.
```sh
cortex engines install llama-cpp -m
```
I think we need better wording here. However, I have to add flag -m to have backward compatibility supported (cortex engines install llama-cpp will behave like before - install latest compatible llama-cpp)

=== 
> 1. Should we align the cortex.llama-cpp version names with [upstream llama.cpp](https://github.com/ggerganov/llama.cpp/releases) eg 0.1.37-b4033 instead of our date format 0.1.37 to 0.1.37-01.11.24

I'm ok with this. But @hiento09 will need to help on CI part.

> 2. cortex engines update llama-cpp: we should have a way to update engines when new engines are available, and delete the old version?

`cortex engines update llama-cpp` will update the latest version. I tested it with below steps
[...]

PR #1650, opened 2024-11-08
chore: some update for engine management

namchuai:
## Describe Your Changes

- Add indicator when engine version is outdated and can be update in CLI
- Remove other engines version when update
- Some improvements

## Fixes Issues

- #1638 

## Self Checklist

### DIFF
--- a/engine/cli/main.cc
+++ b/engine/cli/main.cc
@@ -171,7 +171,7 @@ int main(int argc, char* argv[]) {
       CLI_LOG(
           "New llama.cpp version available: " << config.latestLlamacppRelease);
       CLI_LOG("To update, run: " << commands::GetCortexBinary()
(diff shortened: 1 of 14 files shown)


engine/cli/main.cc
```diff
@@ -101,6 +101,81 @@ int main(int argc, char* argv[]) {
 
   RemoveBinaryTempFileIfExists();
 
+  std::thread t1([]() {
+    // TODO: namh current we only check for llamacpp. Need to add support for other engine
+    auto should_check_for_latest_llamacpp_version = true;
```
vansangpfiev: I think can move the `should_check` part to outside of thread. Should only create new thread if we should check for new update

vansangpfiev (approved):

PR #1665, opened 2024-11-11
feat: simultaneous download

### LINKED ISSUE #1632: feat: Download model simultaneously via API
### Problem Statement

It would be great to allow users to download models simultaneously via the API. Currently, a POST request to `/models/pull` triggers only one download at a time, causing the second request to wait until the first download is completed. This causes a poor UX when integrating cortex.cpp into Jan.

### Feature Idea

Supporting that might lead to a slight change in the response structure of the `/events`

```json
{
  "task": {
    "id": "tinyllama:gguf",
    "items": [
      {
        "bytes": 668788096,
        "checksum": "N/A",
        "downloadUrl": "https://huggingface.co/cortexso/tinyllama/resolve/gguf/model.gguf",
        "downloadedBytes": 0,
        "id": "model.gguf",
        "localPath": "/home/jan/cortexcpp/models/cortex.so/tinyllama/gguf/model.gguf"
      },
      {
        "bytes": 668788096,
        "checksum": "N/A",
        "downloadUrl": "https://huggingface.co/cortexso/tinyllama/resolve/gguf/model.yml",
        "downloadedBytes": 651208498,
        "id": "model.yml",
        "localPath": "/home/jan/cortexcpp/models/cortex.so/tinyllama/gguf/model.yml"
      }
    ],
    "type": "Model"
  },
  "type": "DownloadUpdated"
}
```

So that there should be an array of tasks instead of a task object. This could affect Jan later, so we should update accordingly.

namchuai:
## Describe Your Changes

- Download in parallel
- Fixes: #1632 

## Self Checklist

### DIFF
--- a/engine/services/download_service.cc
+++ b/engine/services/download_service.cc
@@ -356,6 +356,10 @@ void DownloadService::ProcessTask(DownloadTask& task, int worker_id) {
     // if the download has error, we are not run the callback
     ExecuteCallback(task);
     EmitTaskCompleted(task.id);
+    {
+      std::lock_guard<std::mutex> lock(event_emit_map_mutex);
+      event_emit_map_.erase(task.id);
+    }
   }
 
   worker_data->downloading_data_map.clear();
@@ -372,11 +376,19 @@ cpp::result<void, std::string> DownloadService::ProcessMultiDownload(
     auto result = ProcessCompletedTransfers(multi_handle);
     if (result.has_error()) {
       EmitTaskError(task.id);
+      {
+        std::lock_guard<std::mutex> lock(event_emit_map_mutex);
+        event_emit_map_.erase(task.id);
+      }
       return cpp::fail(result.error());
     }
 
     if (task.status == DownloadTask::Status::Cancelled || stop_flag_) {
       EmitTaskStopped(task.id);
+      {
+        std::lock_guard<std::mutex> lock(event_emit_map_mutex);
+        event_emit_map_.erase(task.id);
+      }
       return cpp::fail("Task " + task.id + " cancelled");
     }
   } while (still_running);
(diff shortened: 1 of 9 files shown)


engine/services/download_service.cc
```diff
@@ -179,195 +215,229 @@ cpp::result<bool, std::string> DownloadService::Download(
...
+  // First try to cancel in queue
+  CTL_INF("Stopping task: " << task_id);
+  auto cancelled = task_queue_.cancelTask(task_id);
+  if (cancelled) {
+    EmitTaskStopped(task_id);
+    return task_id;
+  }
+  CTL_INF("Not found in pending task, try to find task " + task_id +
+          " in active tasks");
+  // Check if task is currently being processed
```
vansangpfiev, line 230: nit: should add lock_guard in a block scope

engine/services/download_service.cc
```diff
@@ -382,6 +452,44 @@ void DownloadService::RemoveTaskFromStopList(const std::string& task_id) {
...
+void DownloadService::EmitTaskStopped(const std::string& task_id) {
+  if (auto it = active_tasks_.find(task_id); it != active_tasks_.end()) {
+    event_queue_->enqueue(
+        EventType::DownloadEvent,
+        DownloadEvent{.type_ = DownloadEventType::DownloadStopped,
+                      .download_task_ = *it->second});
+  }
+}
+
+void DownloadService::EmitTaskError(const std::string& task_id) {
```
vansangpfiev, line 471: Why don’t we need lock_guard here but ‘task completed’ we need?

vansangpfiev (approved):
Great work!

engine/services/download_service.cc
namchuai: I think it's won't be necessary because I'm not needed to update downloadedBytes of the item.

engine/services/download_service.cc
namchuai: I will address this in following PR

PR #1705, opened 2024-11-19
feat: add file watcher service

### LINKED ISSUE #1683: epic: Cortex track delete models on file system
### Cortex version

1.0.2

### Describe the issue and expected behaviour

- Problem: Jan users are used to deleting the model folder, rather than deleting on Jan UI
- Users cannot just remove the /models folder as the backend has to update cortex.db
- We should track the file system and update cortex.db so that no action is required for them

Approaches (@louis-jan & @namchuai to add on)
- Build cortex.db as a quick lookup table, but like caching, we still monitor folder changes for better performance.
- Scan it once on launch in the background
- To propose filescan interval
- Alternative: `model has been deleted! -> delete in cortex.db`

## QA:
- Delete model folder (or specifically model.yml file)
- expected result: cortex models list does not list the model

namchuai:
## Describe Your Changes

- Add file watcher service

## Fixes Issues

- fixes #1683 

## Self Checklist

### DIFF
--- a/engine/services/file_watcher_service.h
+++ b/engine/services/file_watcher_service.h
@@ -22,7 +22,7 @@
 class FileWatcherService {
  private:
 #if defined(_WIN32)
-  HANDLE dir_handle = INVALID_HANDLE_VALUE
+  HANDLE dir_handle = INVALID_HANDLE_VALUE;
   HANDLE stop_event;
 #elif defined(__APPLE__)
   FSEventStreamRef event_stream;
--- a/engine/services/file_watcher_service.h
+++ b/engine/services/file_watcher_service.h
@@ -1,5 +1,4 @@
 #include <atomic>
-#include <iostream>
 #include <memory>
 #include <string>
 #include <thread>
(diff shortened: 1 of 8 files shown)


engine/services/file_watcher_service.h
```diff
@@ -0,0 +1,239 @@
...
+#endif
+    running = false;
+    if (watchThread.joinable()) {
+      watchThread.join();
+    }
+    CTL_INF("FileWatcherService stopped!");
+  }
+
+ private:
+  std::string watchPath;
```
vansangpfiev: Maybe should rename to `watch_path` for coding convention consistent.

engine/services/file_watcher_service.h
```diff
@@ -0,0 +1,239 @@
...
+    fd = inotify_init();
+    if (fd < 0) {
+      throw std::runtime_error("Failed to initialize inotify");
+    }
+
+    // Add initial watch on the main directory
+    AddWatch(watchPath);
+
+    char buffer[4096];
+    while (running) {
```
vansangpfiev: Should we put a sleep in the while loop?

engine/services/file_watcher_service.h
```diff
@@ -0,0 +1,239 @@
...
+    watchThread = std::thread(&FileWatcherService::watcherThread, this);
+  }
+
+  void stop() {
+#ifdef _WIN32
+    CloseHandle(dir_handle);
+#endif
+
+#ifdef Linux
+    CleanupWatches();
```
vansangpfiev, line 102: I think we might have a race condition for `watchDescriptors`

engine/services/file_watcher_service.h
namchuai: `read()` is a blocking call. I don't think we should add sleep in there.

vansangpfiev (approved):

PR #1718, opened 2024-11-24
Fix the model in the quick  start API example

grzegorz-bielski:
Hello !

I noticed that the API example in the quick start was a little confusing: 
- it listed the `model` field twice
- it used `mistral` which wasn't pulled in the previous instruction and it didn't use the tag (like `:8b-gguf`)


gabrielle-ong (approved):

gabrielle-ong:
Thank you so much @grzegorz-bielski! Appreciate it!

PR #1720, opened 2024-11-25
feat: add messages api

namchuai:
## Describe Your Changes

- Adding messages API which compatible to https://platform.openai.com/docs/api-reference/messages

## Self Checklist


engine/common/api-dto/messages/delete_message_response.h
```diff
@@ -0,0 +1,19 @@
+#pragma once
+
+#include "common/json_serializable.h"
+
+namespace ApiResponseDto {
```
vansangpfiev: I think we use snake_case for namespace

engine/common/message.h
```diff
@@ -0,0 +1,204 @@
...
+  // The Unix timestamp (in seconds) for when the message was completed.
+  std::optional<uint32_t> completed_at;
+
+  // The Unix timestamp (in seconds) for when the message was marked as incomplete.
+  std::optional<uint32_t> incomplete_at;
+
+  Role role;
+
+  // The content of the message in array of text and/or images.
+  std::vector<std::unique_ptr<Content>> content;
```
vansangpfiev, line 61: Not sure why do we need `unique_ptr`?

namchuai: because `Content` is a pure virtual class. c++ STL containers are not able to handle polymorphic

vansangpfiev:
Great work!
It would be nice if we have unit tests for `ToJson` functions.
Also, I think can consider moving the implementation to the cc files to reduce the compilation time, especially on Windows.

vansangpfiev (approved):

namchuai:
I will add the test in the following PR

PR #1721, opened 2024-11-25
fix: manually add http date header for windows

vansangpfiev:
## Describe Your Changes

- On Windows we set locale to local when convert wstring to std::string. This impacts date field in drogon http header, should add the date as specified in RFC 1123.

## Fixes Issues

- https://github.com/janhq/cortex.cpp/issues/1722

## Self Checklist

### DIFF
--- a/engine/main.cc
+++ b/engine/main.cc
@@ -150,6 +150,8 @@ void RunServer(std::optional<int> port, bool ignore_cout) {
   LOG_INFO << "Please load your model";
 #ifndef _WIN32
   drogon::app().enableReusePort();
+#else
+  drogon::app().enableDateHeader(false);
 #endif
   drogon::app().addListener(config.apiServerHost,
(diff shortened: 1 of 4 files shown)


engine/main.cc
```diff
@@ -150,6 +150,8 @@ void RunServer(std::optional<int> port, bool ignore_cout) {
   LOG_INFO << "Please load your model";
 #ifndef _WIN32
   drogon::app().enableReusePort();
+#else
+  drogon::app().enableDateHeader(false);
```
namchuai, line 154: just wonder, if we send back to client RFC1123 date, then why we need to turn off DateHeader?

vansangpfiev: The date field is not unique in drogon header. If we do not turn off DateHeader we will have 2 date fields.

namchuai (approved):
✅ thank you!

PR #1765, opened 2024-12-03
fix: stop inflight chat completion

vansangpfiev:
## Describe Your Changes

- Signal to engine to stop inflight chat completion if it has request from client.

## Fixes Issues

- https://github.com/janhq/jan/issues/4202

## Self Checklist


engine/controllers/server.cc
```diff
@@ -27,6 +28,13 @@ void server::ChatCompletion(
   LOG_DEBUG << "Start chat completion";
   auto json_body = req->getJsonObject();
   bool is_stream = (*json_body).get("stream", false).asBool();
+  std::string model_id = (*json_body).get("model", "invalid_model").asString();
```
namchuai: should use `auto`?

engine/controllers/server.cc
```diff
@@ -27,6 +28,13 @@ void server::ChatCompletion(
...
+  std::string engine_type;
```
namchuai: I think this is easier to read. Even better if we can make `engine_type` `auto`
```cpp
  std::string engine_type{kLlamaRepo};
  if (inference_svc_->HasFieldInReq(json_body, "engine")) {
    engine_type = (*(json_body)).get("engine", kLlamaRepo).asString();
  }
```

engine/services/inference_service.cc
```diff
@@ -283,6 +293,25 @@ InferResult InferenceService::FineTuning(
...
+bool InferenceService::StopInferencing(const std::string& engine_name,
+                                       const std::string& model_id) {
+  CTL_DBG("Stop inferencing");
+  auto engine_result = engine_service_->GetLoadedEngine(engine_name);
+  if (engine_result.has_error()) {
+    LOG_WARN << "Engine is not loaded yet";
+    return false;
+  }
+
+  if (std::holds_alternative<EngineI*>(engine_result.value())) {
```
namchuai, line 283: hmm, so we don't support stop inference for remote model at the moment?

engine/services/inference_service.cc
```diff
@@ -215,14 +215,24 @@ InferResult InferenceService::GetModels(
...
   for (const auto& loaded_engine : loaded_engines) {
-    auto e = std::get<EngineI*>(loaded_engine);
-    if (e->IsSupported("GetModels")) {
-      e->GetModels(json_body,
-                   [&resp_data](Json::Value status, Json::Value res) {
-                     for (auto r : res["data"]) {
-                       resp_data.append(r);
-                     }
-                   });
+    if (std::holds_alternative<EngineI*>(loaded_engine)) {
```
namchuai, line 206: should we create a single lambda and feed it in `GetModels`

namchuai (approved):
Overally looks good to me. Approved! ✅

engine/services/inference_service.cc
vansangpfiev: yeah, will support it in another PR.

PR #1767, opened 2024-12-04
feat: add thread

namchuai:
## Describe Your Changes

- Add thread APIs.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist


engine/services/thread_service.h
```diff
@@ -0,0 +1,36 @@
...
+#include "common/variant_map.h"
+#include "utils/result.hpp"
+
+class ThreadService {
+ public:
+  explicit ThreadService(std::shared_ptr<ThreadRepository> thread_repository)
+      : thread_repository_{thread_repository} {}
+
+  cpp::result<OpenAi::Thread, std::string> CreateThread(
+      std::optional<std::unique_ptr<OpenAi::ThreadToolResources>>
```
vansangpfiev: Why do we need optional? I think can pass std::nullptr if we use unique_ptr?

engine/repositories/thread_fs_repository.h
```diff
@@ -0,0 +1,62 @@
...
+  /**
+   * Read the thread file and parse to Thread from the file system.
+   */
+  cpp::result<OpenAi::Thread, std::string> LoadThread(
+      const std::string& thread_id) const;
+
+  cpp::result<void, std::string> SaveThread(OpenAi::Thread& thread);
+
+ public:
+  explicit ThreadFsRepository(std::filesystem::path data_folder_path)
```
vansangpfiev: I think can pass by const reference

engine/repositories/thread_fs_repository.h
```diff
@@ -0,0 +1,62 @@
...
+
+  std::filesystem::path GetThreadPath(const std::string& thread_id) const;
+
+  /**
+   * Read the thread file and parse to Thread from the file system.
+   */
+  cpp::result<OpenAi::Thread, std::string> LoadThread(
+      const std::string& thread_id) const;
+
+  cpp::result<void, std::string> SaveThread(OpenAi::Thread& thread);
```
vansangpfiev, line 34: Seems like can use const& for thread parameter?

vansangpfiev (approved):
Great work! Only left some nitpick.

engine/services/thread_service.h
namchuai: ah yeah, my mistake

engine/repositories/thread_fs_repository.h
namchuai: because in the implementation we will call a function `.ToJson`. In order to make it as const, we will need to make `ToJson` const which introduce a lot of changes. I will update it later on.

PR #1784, opened 2024-12-10
Feat/python engine

nguyenhoangthuan99:
## Describe Your Changes

- Cortex supports python engines

## Fixes Issues

- Closes https://github.com/janhq/jan-internal/issues/8


engine/utils/curl_utils.h
```diff
@@ -32,8 +32,9 @@ cpp::result<YAML::Node, std::string> ReadRemoteYaml(const std::string& url);
  *
  * [timeout] is an optional parameter that specifies the timeout for the request. In second.
  */
-cpp::result<Json::Value, std::string> SimpleGetJson(const std::string& url,
-                                                    const int timeout = -1);
+cpp::result<Json::Value, std::string> SimpleGetJson(
+    const std::string& url, const int timeout = -1,
+    std::optional<bool> recursive = true);
```
namchuai: why not just `bool recursive = true` if needed?

engine/services/download_service.cc
```diff
@@ -374,6 +374,9 @@ void DownloadService::ProcessTask(DownloadTask& task, int worker_id) {
       CTL_ERR("Failed to init curl!");
       return;
     }
+    if (!std::filesystem::exists(item.localPath.parent_path())) {
```
namchuai: I think the component which use download service will need to ensure the parent folder to be exists. Download service should not worry about create directory.

engine/test/components/test_remote_engine.cc
```diff
@@ -42,7 +42,7 @@ TEST_F(RemoteEngineTest, OpenAiToAnthropicRequest) {
 
     auto data = json_helper::ParseJsonString(message_with_system);
 
-    remote_engine::TemplateRenderer rdr;
+    extensions::TemplateRenderer rdr;
```
namchuai, line 45: any problem if we using `jinja::RenderTemplate`?

engine/common/base.h
```diff
@@ -46,3 +46,16 @@ class BaseEmbedding {
 
   // The derived class can also override other methods if needed
 };
+
+class BasePythonModel {
+ public:
+  virtual ~BasePythonModel() {}
```
namchuai: `virtual ~BasePythonModel = default` is a bit better

engine/common/base.h
```diff
@@ -46,3 +46,16 @@ class BaseEmbedding {
...
+
+  // Model management
+  virtual void Inference(
+      const HttpRequestPtr& req,
+      std::function<void(const HttpResponsePtr&)>&& callback) = 0;
```
namchuai: Having Drogon's dependencies here is against Clean architecture. Models/entities should not be depends on framework

engine/controllers/models.cc
```diff
@@ -282,7 +292,8 @@ void Models::GetModel(const HttpRequestPtr& req,
       auto resp = cortex_utils::CreateCortexHttpTextAsJsonResponse(ret);
       resp->setStatusCode(drogon::k200OK);
       callback(resp);
-    } else {
+    } else if (model_config.engine == kOpenAiEngine ||
```
namchuai: If we go down this path, I think we should check engine == python here, rather than check engine == any remote engine. Because remote engine will have a lot more in the future (mistral, etc.)

so, I think in else if case, should be kPython, leave the remote engine to else as before

engine/controllers/server.h
```diff
@@ -25,7 +25,8 @@ namespace inferences {
 class server : public drogon::HttpController<server, false>,
                public BaseModel,
                public BaseChatCompletion,
-               public BaseEmbedding {
+               public BaseEmbedding,
+               public BasePythonModel {
```
namchuai: Maybe I'm missing something, I'm not sure why do we need to create interface like BaseModel, BasePythonModel for this server controller?

engine/extensions/python-engine/python_engine.cc
```diff
@@ -0,0 +1,919 @@
...
+constexpr const int kFileLoggerOption = 0;
+
+static size_t WriteCallback(char* ptr, size_t size, size_t nmemb,
+                            std::string* data) {
+  data->append(ptr, size * nmemb);
+  return size * nmemb;
+}
+
+PythonEngine::PythonEngine() {
+  curl_global_init(CURL_GLOBAL_ALL);
```
namchuai: `curl_global_init` is already invoked at main.cc. I think we can remove this.

engine/extensions/python-engine/python_engine.cc
```diff
@@ -0,0 +1,919 @@
...
+    response.error_message = curl_easy_strerror(res);
+  } else {
+    response.body = response_string;
+  }
+
+  curl_slist_free_all(headers);
+  curl_easy_cleanup(curl);
+  return response;
+}
+CurlResponse PythonEngine::MakeDeleteRequest(const std::string& model,
```
namchuai, line 185: can we try to reuse `curl_utils` instead?

engine/utils/curl_utils.cc
```diff
@@ -242,8 +242,8 @@ cpp::result<YAML::Node, std::string> ReadRemoteYaml(const std::string& url) {
   }
 }
 
-cpp::result<Json::Value, std::string> SimpleGetJson(const std::string& url,
-                                                    const int timeout) {
+cpp::result<Json::Value, std::string> SimpleGetJson(
```
namchuai: I'm not sure if we adding `recursive` here. The util should just be a simple GET. The user of this API should be in charge to determine if they need to recursive get another API.

And I can't find where in this PR we use this?

namchuai (approved):

PR #1790, opened 2024-12-11
fix: load engine linux

namchuai:
## Describe Your Changes

- Fix load engine issue on linux

## Self Checklist

### DIFF
--- a/engine/cli/commands/engine_list_cmd.cc
+++ b/engine/cli/commands/engine_list_cmd.cc
@@ -13,7 +13,6 @@
 // clang-format on
 
 namespace commands {
-
 bool EngineListCmd::Exec(const std::string& host, int port) {
   // Start server if server is not started yet
   if (!commands::IsServerAlive(host, port)) {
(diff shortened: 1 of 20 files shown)


engine/cli/commands/engine_list_cmd.cc
```diff
@@ -32,21 +31,17 @@ bool EngineListCmd::Exec(const std::string& host, int port) {
       .host = host + ":" + std::to_string(port),
       .pathParams = {"v1", "engines"},
   };
+  CTL_INF("NamH URL: " << url.ToFullPath());
```
vansangpfiev: Should remove debug log

engine/services/engine_service.cc
```diff
@@ -753,27 +772,32 @@ void EngineService::RegisterEngineLibPath() {
...
+      paths.push_back(std::move(cuda_path));
+      paths.push_back(std::move(engine_dir_path));
+
+      CTL_DBG("Registering dylib for "
+              << ne << " with " << std::to_string(paths.size()) << " paths.");
+      for (const auto& path : paths) {
+        CTL_DBG("Registering path: " << path.string());
+      }
+
+      auto reg_result = dylib_path_manager_->RegisterPath(ne, paths);
```
vansangpfiev, line 794: Seems like we register twice for Windows.
For Linux, I think we don't need to register here, also the LD_LIBRARY_PATH does not work with running process. We have already registered with `server_start_cmd`

vansangpfiev (approved):

PR #1798, opened 2024-12-16
fix: add support image url for jan

namchuai:
## Describe Your Changes

- Add support for image url jan.
```json
"content": [
                {
                    "text": {
                        "annotations": [],
                        "value": "What is this image about?"
                    },
                    "type": "text"
                },
                {
                    "image_url": {
                        "detail": "auto",
                        "url": "threads/jan_1734081137/files/01JEZMN5DKCX8MJG2W7P6NYCC0.png"
                    },
                    "type": "image_url"
                }
            ],
```

## Self Checklist

### DIFF
--- a/engine/common/message_content_image_url.h
+++ b/engine/common/message_content_image_url.h
@@ -4,14 +4,21 @@
 
 namespace OpenAi {
 
-struct ImageUrl {
-  // The external URL of the image, must be a supported image types: jpeg, jpg, png, gif, webp.
+struct ImageUrl : public JsonSerializable {
+  /**
+   * The external URL of the image, must be a supported image types:
+   * jpeg, jpg, png, gif, webp.
+   */
   std::string url;
 
-  // Specifies the detail level of the image. low uses fewer tokens, you can opt in to high resolution using high. Default value is auto
+  /**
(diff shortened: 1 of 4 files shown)


engine/common/message_content_image_url.h
```diff
@@ -38,18 +57,19 @@ struct ImageUrlContent : Content {
...
     }
 
     try {
-      ImageUrlContent content{"image_url"};
-      ImageUrl image_url;
-      image_url.url = std::move(json["image_url"]["url"].asString());
-      image_url.detail = std::move(json["image_url"]["detail"].asString());
-      content.image_url = std::move(image_url);
+      auto image_url =
+          ImageUrl(std::move(json["image_url"]["url"].asString()),
```
vansangpfiev: I think we don't need `std:move` here since we have const ref for parameters:
```
ImageUrl(const std::string& url, const std::string& detail = "auto")
      : url{url}, detail{detail} {}
```

vansangpfiev (approved):
LGTM overall.

engine/common/message_content_image_url.h
namchuai: ah yeah, let me remove the `std::move`

PR #1814, opened 2024-12-19
feat: rendering chat_template

namchuai:
## Describe Your Changes

-

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/engine/services/model_service.cc
+++ b/engine/services/model_service.cc
@@ -9,10 +9,11 @@
 #include "config/yaml_config.h"
 #include "database/models.h"
 #include "hardware_service.h"
+#include "services/inference_service.h"
 #include "utils/cli_selection_utils.h"
-#include "utils/cortex_utils.h"
 #include "utils/engine_constants.h"
 #include "utils/file_manager_utils.h"
+#include "utils/gguf_metadata_reader.h"
 #include "utils/huggingface_utils.h"
 #include "utils/logging_utils.h"
 #include "utils/result.hpp"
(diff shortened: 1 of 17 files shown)


engine/services/model_service.cc
```diff
@@ -1208,3 +1223,39 @@ ModelService::MayFallbackToCpu(const std::string& model_path, int ngl,
...
+      file_manager_utils::ToAbsoluteCortexDataPath(
+          std::filesystem::path(file)));
+  if (!model_metadata_res.has_value()) {
+    CTL_ERR("Failed to read metadata: " + model_metadata_res.error());
+    return cpp::fail("Failed to read metadata: " + model_metadata_res.error());
+  }
+  return std::move(*model_metadata_res);
+}
+
+std::optional<std::shared_ptr<ModelMetadata>>
```
vansangpfiev: I think can use std::shared_ptr and return nullptr here

engine/utils/jinja_utils.h
```diff
@@ -0,0 +1,27 @@
+#pragma once
+
+#include <json/value.h>
+#include <string>
+
+#include "extensions/remote-engine/template_renderer.h"
+#include "utils/chat-template.hpp"
+#include "utils/result.hpp"
+
+namespace jinja {
```
vansangpfiev, line 10: I think we use `inja` not `jinja`

vansangpfiev (approved):

engine/utils/jinja_utils.h
namchuai: I think jinja is more accurate in semantic. inja is a library tool.

engine/services/model_service.cc
namchuai: ah yeah, thank you, let me update it.

PR #1826, opened 2024-12-24
feat: add assistant api

namchuai:
## Describe Your Changes

- Add assistant API

## Self Checklist


engine/common/assistant.h
```diff
@@ -126,8 +172,7 @@ struct Assistant {
    * requires a list of file IDs, while the file_search tool requires a list
    * of vector store IDs.
    */
-  std::optional<std::variant<ThreadCodeInterpreter, ThreadFileSearch>>
-      tool_resources;
+  std::optional<std::unique_ptr<OpenAi::ToolResources>> tool_resources;
```
vansangpfiev: I think we should use `std::unique_ptr` instead of `std::optional<std::unique_ptr>`

engine/common/assistant_code_interpreter_tool.h
```diff
@@ -0,0 +1,33 @@
...
+  AssistantCodeInterpreterTool(AssistantCodeInterpreterTool&&) = default;
+
+  AssistantCodeInterpreterTool& operator=(AssistantCodeInterpreterTool&&) =
+      default;
+
+  ~AssistantCodeInterpreterTool() = default;
+
+  static cpp::result<AssistantCodeInterpreterTool, std::string> FromJson(
+      const Json::Value& json) {
+    AssistantCodeInterpreterTool tool;
```
vansangpfiev, line 22: Just curious, why is `json` unused here?

engine/repositories/assistant_fs_repository.cc
```diff
@@ -0,0 +1,207 @@
...
+  std::unique_lock lock(GrabAssistantMutex(assistant.id));
+  auto path = GetAssistantPath(assistant.id);
+
+  if (std::filesystem::exists(path)) {
+    return cpp::fail("Assistant already exists: " + assistant.id);
+  }
+
+  std::filesystem::create_directories(path);
+  auto assistant_file_path = path / kAssistantFileName;
+  std::ofstream assistant_file(assistant_file_path);
```
vansangpfiev: Is it necessary since we create file if it does not exist in `SaveAssistant` function?

vansangpfiev (approved):

engine/repositories/assistant_fs_repository.cc
namchuai: It's needed because we need to create `assistant.json` inside the folder. ModifyAssistant also use `SaveAssistant` so can't move the create `assistant.json` there.

engine/common/assistant_code_interpreter_tool.h
namchuai: for this one, it's not needed. let me remove it.

PR #1891, opened 2025-01-27
Docs improvements

ramonpzg:
# General Documentation Improvements

## General Improvements
- fixed typos and misspellings
- normalized saying Cortex in favor of Cortex.cpp
- simplified the repo's main README
- added images where possible to illustrate the topic being discussed in a section, for example, to show the way in which `cortexso` hub saves model versions in different git branches, or how to query the various tables in our cortex.db sqlite file
- Pedantic changes:
  - favor use and using instead of utilizing
  - The AI space is moving so fast that it is best to rephrase "industry standard" to "sensible defaults" or something similar
  - Removed words such as simple, super easy, and extremely easy.
- Broke long lines into smaller ones
- Separated code block where the command and the output were on the same line. This makes it easier for readers to copy the command they want to use without the response they don't need.
- Remove `>` from in front of many shell commands across the docs. This adds friction between users copying the command with one click and having to move to the beginning of the line to delete the greater than sign `>`.
- remove TensorRT-LLM from the docs
- improved python engine explanation
- Removed warning where appropriate

## Noticeable changes
- Simplified explanations, where possible
- Created a Guide for text generation
- Re-created the structured outputs section


dan-menlo (approved):
Lgtm

PR #1919, opened 2025-02-03
feat: Add 'Size' column for engines in the list command

github-roushan:
## Describe Your Changes

- This PR introduces a "Size" column in the Engines list sub command. 
- While the sizes of models are visible in the Cortex API reference when accessing web URLs (e.g., http://localhost:39281/#tag/engines/GET/v1/engines/llama-cpp/releases/v0.1.49 ), this information was previously missing in the Command Line Utility. This update adds the missing "Size" column to the command-line interface.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

### DIFF
--- a/engine/utils/cli_selection_utils.h
+++ b/engine/utils/cli_selection_utils.h
@@ -25,12 +25,24 @@ inline void PrintMenu(
   std::endl(std::cout);
 }
 
+inline std::optional<int> GetNumericValue(const std::string& sval) {
+  try {
+      return std::stoi(sval);
+  } catch (const std::invalid_argument&) {
+      // Not a valid number
+      return std::nullopt; 
+  } catch (const std::out_of_range&) {
+      // Number out of range
+      return std::nullopt;
+  }
+}
+
 inline std::optional<std::string> PrintModelSelection(
     const std::vector<std::string>& downloaded,
     const std::vector<std::string>& availables,
     const std::optional<std::string> default_selection = std::nullopt) {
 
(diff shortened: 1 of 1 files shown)


github-roushan:
CLI output before
Screenshot from 2025-02-04 02-02-57


CLI Output After:
Screenshot from 2025-02-04 02-00-51

engine/cli/commands/engine_list_cmd.cc
```diff
@@ -69,6 +69,24 @@ bool EngineListCmd::Exec(const std::string& host, int port) {
         selected_variant_result.value()["version"].asString());
   }
 
+  std::unordered_map<std::string, std::string> variantSizeMapping;
```
vansangpfiev: We use snake_case for variables. It would be nice if `variantSizeMapping` and `releaseData` were in the snake case format as well.

github-roushan: Done Fixes in the latest commit in this PR. Please check

ohaiibuzzle:
Isn't this just listing the size as downloaded from GitHub? 

This command (`cortex engines list`) lists the installed engines on-device, not the GitHub downloads, so it would make sense if it shows the actual space taken up by them and not what the compressed download sizes.

This way it is showing a ~4.5MB engine installed on my systems as only 1.5MB (compressed inside a zip)

github-roushan:
Hey @ohaiibuzzle . Modified the code to include both download size and size on system

image

PR #1925, opened 2025-02-04
feat: QoL improvements for server when running standalone

ohaiibuzzle:
## Describe Your Changes
This mainly adds some nice QoL feature for Cortex Server when it is being ran on its own (ala. directly executing the `cortex-server` binary):
- Include a help output on `-h` or `--help` for the available options
- Web server now run in a separate detached thread instead of blocking the main thread.
- Proper SIGINT handling for gracefully shutting down the web server.

## Fixes Issues

- None, this is just pure QoL.

## Self Checklist

### DIFF
--- a/engine/main.cc
+++ b/engine/main.cc
@@ -1,6 +1,8 @@
 #include <drogon/HttpAppFramework.h>
 #include <drogon/drogon.h>
+#include <trantor/utils/Logger.h>
 #include <memory>
+#include <mutex>
 #include "controllers/assistants.h"
 #include "controllers/configs.h"
 #include "controllers/engines.h"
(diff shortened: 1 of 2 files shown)


louis-jan:
Awesome!

vansangpfiev:
Hi @ohaiibuzzle, we encountered an issue before with `Ctrl + C` on Windows when running cortex CLI: it also terminated the cortex Server process. A workaround for this is to ignore `Ctrl + C` event on the Server side and use `/destroy` endpoint to stop the Server (yes - it is tricky and ugly).

Can you please help to check the PR on the Windows?

ohaiibuzzle:
@vansangpfiev I will check for that later, I only tested on macOS and Linux.

Although, afaik, in Windows console properties are shared by all processes sharing that same console host instance, so what likely happened is when that Ctrl handler is installed, it becomes global for the console host instance shared by cortex and cortex-server.

So even when cortex-server moved to background it still maintains itself as the Ctrl C handler.

As for the solution: This on the CLI *should* work: https://stackoverflow.com/a/71364777

ohaiibuzzle:
~~Update: I compiled it on my Windows system. Could not reproduce. Even Ctrl+C  have the server instance starts up correctly in the background. Perhaps I am missing something...?~~

Yes, I managed to reproduce it. Fixed by setting `CREATE_NO_WINDOW` to CreateProcess which detaches the server child completely from the parent CLI: https://learn.microsoft.com/en-us/windows/win32/procthread/process-creation-flags

engine/main.cc
```diff
@@ -276,11 +297,37 @@ void RunServer(std::optional<std::string> host, std::optional<int> port,
...
+    drogon::app().run();
+    shutdown_signal = true;
+  }).detach();
+
+  // Now this thread can monitor the shutdown signal
+  while (!shutdown_signal) {
+    std::this_thread::sleep_for(std::chrono::seconds(1));
+  }
+
   if (hw_service->ShouldRestart()) {
```
vansangpfiev, line 311: I think this check should be moved before the while loop

vansangpfiev (approved):

PR #1951, opened 2025-02-11
doc: Update build from source documentation

### LINKED ISSUE #1950: doc: Build from source instruction does not work
On my Mac M1, the following instructions do not work

```
git clone https://github.com/janhq/cortex.cpp
cd cortex.cpp
git submodule update --init --recursive
cd engine/vcpkg && ./bootstrap-vcpkg.sh
cd ../build && cmake .. && make -j4
```

One obvious bug is that `build/` directory does not exist, and it must be created before the last step. After creating `build/` directory, I faced with missing `jsoncpp`

```
CMake Error at cli/CMakeLists.txt:62 (find_package):
  Could not find a package configuration file provided by "jsoncpp" with any
  of the following names:

    jsoncppConfig.cmake
    jsoncpp-config.cmake

  Add the installation prefix of "jsoncpp" to CMAKE_PREFIX_PATH or set
  "jsoncpp_DIR" to a directory containing one of the above files.  If
  "jsoncpp" provides a separate development package or SDK, be sure it has
  been installed.
```

At a glance into our Macos build CI https://github.com/janhq/cortex.cpp/blob/dev/.github/workflows/template-build-macos.yml, there are some differences in the command used. Can we update the correct instructions, and possibly considate the build commands for README and CI, so that there is a single source of truth?

gau-nernst:
## Describe Your Changes

This PR brings back the old Build from source instructions from 9f5a7968. To keep the main README clutter-free, it's been put under `BUILDING.md`

Linux and Macos instructions are combined since they are the same

## Fixes Issues

- Closes #1950

## Self Checklist

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -1,0 +1,102 @@
+# Build Cortex.cpp from source
+
+Firstly, clone the Cortex.cpp repository [here](https://github.com/janhq/cortex.cpp) and initialize the submodules:
+
+```bash
+git clone https://github.com/janhq/cortex.cpp
+cd cortex.cpp
+git submodule update --init --recursive
+```
+
+#### Windows
+
+1. Navigate to the `engine` folder.
+2. Configure the vpkg:
+
+```bash
+cd vcpkg
+./bootstrap-vcpkg.bat
+vcpkg install
+```
+
+3. Build the Cortex.cpp inside the `engine/build` folder (you can change `-DCMAKE_TOOLCHAIN_FILE` to use your own `vcpkg`):
+
+```bash
+mkdir build
+cd build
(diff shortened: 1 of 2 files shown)


vansangpfiev (approved):

gau-nernst:
When I tried this on my Ubuntu machine, `cmake` is not available -> I had to run `sudo apt install cmake`. My Macbook already has `cmake` from homebrew, hence I didn't encounter this error.

To further improve the doc, perhaps we can't assume that users have `cmake` installed. Should we add a section on install cmake?

vansangpfiev:
> When I tried this on my Ubuntu machine, `cmake` is not available -> I had to run `sudo apt install cmake`. My Macbook already has `cmake` from homebrew, hence I didn't encounter this error.
> 
> To further improve the doc, perhaps we can't assume that users have `cmake` installed. Should we add a section on install cmake?

Agree, I think we should add a `prerequisite` section

PR #2017, opened 2025-02-24
fix: append stop words from request in case of using template renderer for local engine

vansangpfiev:
## Describe Your Changes

This pull request includes an important change to the `InferenceService::HandleChatCompletion` method in the `engine/services/inference_service.cc` file. The change ensures that the end-of-sequence (EOS) token is correctly appended to the `stop` field in the JSON body if it is not already present.

Key change:

* [`engine/services/inference_service.cc`](diffhunk://#diff-e165ef8bee4fa9dfd1a2d044abdc345eb7784ee2708a14cd582cb33ec9beceb0R69-R83): Added logic to check if the `stop` field in the JSON body already contains the EOS token and append it if necessary.

## Fixes Issues

- https://github.com/janhq/cortex.cpp/issues/2016

## Self Checklist


qnixsynapse (approved):
LGTM but I would prefer these all things to be handled by llamacpp itself at some point.

PR #2022, opened 2025-02-24
Fix: Include algorithm header file to fix build on other distros

qnixsynapse:
## Describe Your Changes
Build currently breaks on distributions other than Ubuntu because of a missing header file.
This patch attempts to fix that.

### DIFF
--- a/engine/common/download_task_queue.h
+++ b/engine/common/download_task_queue.h
@@ -1,3 +1,4 @@
+#include <algorithm>
 #include <condition_variable>
 #include <deque>
 #include <mutex>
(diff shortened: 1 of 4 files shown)


gau-nernst (approved):

engine/common/download_task_queue.h
```diff
@@ -6,6 +6,7 @@
 #include <string>
 #include <unordered_map>
 #include "common/download_task.h"
+#include <algorithm>
```
vansangpfiev: Please format the file before pushing it. If you're using VS Code, you can use the shortcut `Shift + Alt + F` to format it easily.

vansangpfiev (approved):

engine/common/download_task_queue.h
qnixsynapse: I use neovim... Should I do it for every file?

nevermind, did it for every file in this change.

PR #2057, opened 2025-03-03
Add a timeout and disable compression to prevent libcurl OOM errors

qnixsynapse:
## Describe Your Changes
While downloading stuff though cortex, libcurl was running OOM, so disabled compression and added a timeout to try to prevent such errors

edit: I will fix the formatting later.


qnixsynapse:
Closing... Probably issue is with something else.

PR #2072, opened 2025-03-06
chore: address compiler warnings

github-roushan:
## Describe Your Changes
- This pull request addresses a range of compiler warnings across multiple files, improving code quality and maintainability.
- It implements handling for `[[nodiscard]]` attributes where return values were previously being ignored, ensuring proper error checking and resource management.
- Added comprehensive error checking and logging to provide better diagnostics and identify potential issues during runtime.

## Fixes Issues

- Closes #
- Closes #

## Self Checklist

- [x] Added relevant comments, esp in complex areas
- [ ] Updated docs (for bug fixes / features)
- [ ] Created issues for follow-up changes or refactoring needed

PR #2075, opened 2025-03-06
fix: ignored-attributes warning in CommandExecutor

qnixsynapse:
## Describe Your Changes

-The warning was triggered by using `decltype(&PCLOSE)` to deduce the deleter type for the `std::unique_ptr` member `m_pipe`. While `decltype(&PCLOSE)` correctly deduces the function pointer type, the compiler flags this as a potential issue because attributes associated with the function pointer type (if any existed in the definition of `PCLOSE`) are ignored when used as a template argument.
To resolve this warning, the deleter type for `std::unique_ptr` is now explicitly specified as `int (*)(FILE*)`. This directly states the function pointer type expected for `PCLOSE` and avoids the attribute deduction issue.

This is a code quality improvement and does not change the functionality of the `CommandExecutor` class. It only modifies the type declaration of the `m_pipe` member.

### DIFF
--- a/engine/utils/command_executor.h
+++ b/engine/utils/command_executor.h
@@ -20,7 +20,7 @@ class CommandExecutor {
     if (!pipe) {
       throw std::runtime_error("popen() failed!");
     }
-    m_pipe = std::unique_ptr<FILE, decltype(&PCLOSE)>(pipe, PCLOSE);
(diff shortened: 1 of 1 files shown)


github-roushan:
hey @qnixsynapse  a PR for this already open [Pr #2072 ](https://github.com/janhq/cortex.cpp/pull/2072)

PR #2138, opened 2025-03-17
chore: Move to C++17 and fix extra warnings

### LINKED ISSUE #2130: epic: Migrate repo to C++ 17
## Goal 
To have a consistant C++ version for supporting every compiler and platform

## Success Criteria
Build should pass with COTEX_CQA=ON cmake flag

## Key Decisions

Our original decision was to choose between C++ 17 and C++ 20:
- Not all compiler support C++ 20, especially on Linux

Why are we not sticking with what we did before? (hybrid C++ 17 and 20?)
- Compiler would complain if we used C++ 20 in C++ 17
- We want to make it consistent across all platforms
- Having a mono C++17 repo helps us detect memory leaks, corruption issues, SegFaults, dangling pointers etc

When do we anticipate moving to C++ 20?
- When `llama.cpp` moves to C++20
- Linux compiler support sucks
- Key question: do we need C++ 20? (or 23)

Key question: what is the OS we wish to support?
- If we run our own server, we should use the latest C++ version
- Question 1: Raspberry Pi?
- Question 2: Scaled Inference (i.e. efficiency)

## Tasklist
- [x] Change designated initialization to list initialization (which is supported by C++17) 
- [ ] Address and remove unused variables and functions

qnixsynapse:
## Describe Your Changes

- The change migrates the codebase from C++20 to C++17 by replacing C++20's designated initialization with list initialization which is supported C++17 across platforms.
Currently, tested on Linux.

- Also, fixed most compiler warning and removed/added attributes to some unused variables and functions

## Fixes Issues

- Closes #2130 

## Self Checklist


vansangpfiev:
LGTM.
It would be great if we can enable `-Wpedantic -Werror` for linux CI.

qnixsynapse:
Yup. I intend to enable CQA(which contains these flags) on *nix machines with valgrind and ASan.

cc: @david-menloai

vansangpfiev (approved):

PR #2155, opened 2025-03-20
fix: prevent unlimited loop due to invalid filename in path

### LINKED ISSUE #2153: bug: `/v1/files/` invalid filename will hang the server
### Cortex version

dev 2e1dfa02

### Describe the issue and expected behaviour

Something like this will hang the server

```bash
curl http://127.0.0.1:3928/v1/files -X POST -F "data=@libvulkan.so;filename=../hi" -F purpose=assistants
```

Digging further, I find that this loop never exits in such cases https://github.com/menloresearch/cortex.cpp/blob/2e1dfa02ccdf7816b02223f01d95e658c7296fa8/engine/repositories/file_fs_repository.cc#L20-L25

We might want to just sanitize the filename as a **string** instead of a path i.e. no weird characters, including `/`. This will also solve an issue if the filename appears to be a subdirectory

```bash
curl http://127.0.0.1:3928/v1/files -X POST -F "data=@libvulkan.so;filename=abc/hello" -F purpose=assistants
{"message":"Failed to open file for writing: <redacted>/cortexcpp/files/abc/hello"}
```

(or we can just store the binary file in a DB 😆)

### Steps to Reproduce

_No response_

### Screenshots / Logs

_No response_

### What is your OS?

- [ ] Windows
- [ ] Mac Silicon
- [ ] Mac Intel
- [ ] Linux / Ubuntu

- [ ] cortex.llamacpp (default)

### Hardware Specs eg OS version, GPU

_No response_

qnixsynapse:
## Describe Your Changes

- This PR addresses an infinite loop issue in the SanitizePath function. The bug occurs when the user-supplied filename (e.g., ../hi) results in a resolved path that is outside of the base directory. When the resolved path reaches the root directory, the loop fails to exit because the root is its own parent, causing the loop to iterate indefinitely.

## Fixes Issues

- Closes #2153 

## Self Checklist

- [X] Added relevant comments, esp in complex areas

### DIFF
--- a/engine/repositories/file_fs_repository.cc
+++ b/engine/repositories/file_fs_repository.cc
@@ -22,6 +22,9 @@ std::filesystem::path SanitizePath(const std::filesystem::path& user_input,
     if (std::filesystem::equivalent(p, abs_base)) {
       return resolved_path;
     }
+    if (p == p.parent_path()) {  // reached the root directory
+      break;
+    }
   }
   return {};
 }


gau-nernst (approved):
No more hang, though I still think it's better to sanitize the filename string itself (checks including, but not limited to, no `/`), instead of trying to parse it as path and make sure it falls under cortex directory.

Approve to unblock. You can add test if you have time. Thank you!

PR #2207, opened 2025-05-20
fix bug where for local models, delete only the model passed in, not …

ethanova:
…all local models

## Describe Your Changes
In [Jan](https://github.com/menloresearch/jan), I experienced a bug where I would import a model and all of a sudden, all of my other models were gone. I just did some digging finally and it appears that actually what was happening was that I previously deleted a local model, and that deleted all of my local models (all symbolic links). This should provide expected behavior of DeleteModel

## Fixes Issues

- I haven't created a bug, but I think this is the fix

## Self Checklist

- [?] Updated docs (for bug fixes / features)
- [na] Created issues for follow-up changes or refactoring needed


qnixsynapse (approved):

louis-jan (approved):
LGTM

